Issue #982, opened 2025-03-19
Fix the failing export of the `azure-devops` workspace for the `release-1.5` branch.


davidfestal:
The `azure-devops` workspace introduced on branch `release-1.5` by PR #897 failed for plugin `plugins/catalog-backend-module-azure-devops-annotator-processor`.

The workflow error is this one: https://github.com/redhat-developer/rhdh-plugin-export-overlays/actions/runs/13945146703/job/39030562918#step:13:429

and shows that an embedded package (`@backstage-community/plugin-azure-devops-common`) should be added during the export of this plugin.

PR #1030, opened 2025-04-29
chore(ci): add GitHub Action to add stale label for for inactive PRs and issues

josephca:
Fixes https://issues.redhat.com/browse/RHIDP-7091

### DIFF
--- a/.github/workflows/stale.yaml
+++ b/.github/workflows/stale.yaml
@@ -1,0 +1,30 @@
+name: Label stale PRs and issues
+
+on:
+  schedule:
+    - cron: "30 1 * * *" # Runs daily at 1:30 AM UTC
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+
+    permissions:
+      issues: write
+      pull-requests: write
+
+    steps:
+      - uses: actions/stale@v9
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+          days-before-issue-stale: 30
+          days-before-issue-close: -1 # never automatically close stale issues
+          stale-issue-label: "stale"
+          stale-issue-message: "This issue has been automatically marked as stale due to inactivity."
+          exempt-issue-labels: "stale"
+          # close-issue-message: 'Closing this issue due to prolonged inactivity.'
+          days-before-pr-stale: 30
+          days-before-pr-close: -1 # never automatically close stale PRs
+          stale-pr-label: "stale"
+          stale-pr-message: "This pull request has been automatically marked as stale due to inactivity."
+          # close-pr-message: 'Closing this pull request due to prolonged inactivity.'
+          exempt-pr-labels: "stale,help wanted to test"


.github/workflows/stale.yaml
```diff
@@ -0,0 +1,30 @@
...
+          stale-issue-label: "stale"
+          stale-issue-message: "This issue has been automatically marked as stale due to inactivity."
+          exempt-issue-labels: "stale"
+          # close-issue-message: 'Closing this issue due to prolonged inactivity.'
+          days-before-pr-stale: 30
+          days-before-pr-close: -1 # never automatically close stale PRs
+          stale-pr-label: "stale"
+          stale-pr-message: "This pull request has been automatically marked as stale due to inactivity."
+          # close-pr-message: 'Closing this pull request due to prolonged inactivity.'
+          exempt-pr-labels: "stale,help wanted to test"
```
nickboldt, line 30: @davidfestal should the issues with "help wanted" on them be exempt from being marked stale? I'm thinking NO, but maybe you want them to persist more than 30 days without being addressed?

.github/workflows/stale.yaml
```diff
@@ -0,0 +1,30 @@
...
+    permissions:
+      issues: write
+      pull-requests: write
+
+    steps:
+      - uses: actions/stale@v9
+        with:
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+          days-before-issue-stale: 30
+          days-before-issue-close: -1 # never automatically close stale issues
```
nickboldt, line 20: @davidfestal should we mark PRs stale after 14 days of inactivity, and close them after 30 days? Will that mean they get recreated by your crontabs?

davidfestal (commented):
Are we sure we want to start with setting up a mechanism to cleanup old PRs / issues, while we didn't even define issue / PR assignment rules and additional repository permissions for assigned team members ?
I'm all in favor of avoiding old issues / PRs, but it seems to me that, to achieve this goal, we should to start by unlocking wider involvement in fixing / merging them, than by implementing a cleanup mechanism.

So for now I'd prefer avoiding the automatic closing of issues or PRs. Setting the stale label is a good start though, and we can start doing it right now.
This doesn't prevent us revisiting this later, when the processes around the overly repository (PR review assignment, code owners, automatic tests, release branch obsolescence rules, etc ...) are discussed and agreed on.

.github/workflows/stale.yaml
davidfestal: > @davidfestal should we mark _issues_ like #982 stale after 14 days of inactivity, and close them after 30 days? 

I think that the proposed rule of simply marking them stale after 30 days and not closing them is good for now.
Please see my general comment about why: https://github.com/redhat-developer/rhdh-plugin-export-overlays/pull/1030#pullrequestreview-2806532411

> Also why are we using issues here, when we have JIRA? 😕

I created it just to track that this PR was failing, stick the result of my first analysis, and take the opportunity to add a `help wanted` label.
This doesn't prevent us from adding JIRA issues related to the overlay repository.

josephca:
From the discussions/inputs above, it seems the current change should be a good start - by adding 'stale' label for the issues/prs older than 30 days and not closing them. 

@nickboldt @davidfestal - please approve if you agree. Thanks!

nickboldt (approved):
/lgtm 
/approve

PR #1039, opened 2025-05-07
[release-1.5] chore: add first version of a codeowners file

christoph-jerolimov:
Adds a codeowners file with a fallback to @redhat-developer/rhdh-plugins-maintainers and @davidfestal 

Afaik: Users must still have write access. So users who are not part of our org or have write access don't matter.

The last matching pattern takes precedence. See https://help.github.com/articles/about-codeowners/

The initial version is based on https://github.com/backstage/community-plugins/blob/main/.github/CODEOWNERS and https://github.com/redhat-developer/rhdh-plugins/blob/main/.github/CODEOWNERS

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,41 @@
+# This file registers ownership for certain parts of the rhdh-plugin-export-overlays repo.
+# Review from a member of the corresponding code owner is required to merge pull requests.
+#
+# The last matching pattern takes precedence.
+# https://help.github.com/articles/about-codeowners/
+#
+# The list is alphabetical ordered.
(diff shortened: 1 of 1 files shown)


nickboldt:
LGTM but with the caveat that a lot of the people listed (and the group) don't already have access to this repo.

image

So as a prereq I guess someone has to add them as writers in this repo, otherwise ...

https://imgflip.com/i/9ta0ja

:D

christoph-jerolimov:
CODEOWNERS file is now green. I've invited some people to the team and I've removed everything that it not working for now.

davidfestal (approved):

davidfestal:
Do we also want to add that on the [template-for-releases](https://github.com/redhat-developer/rhdh-plugin-export-overlays/tree/template-for-releases) branch ?

PR #1135, opened 2025-05-12
Support patterns to identify active release branches...

davidfestal:
Support patterns to identify active release branches and remove the old `release 1.4` branch from watched branches.


nickboldt:
Seems like this was related to https://issues.redhat.com/browse/RHIDP-7507 and merged without review. Might have been good to add future branch patterns so we don't have to maintain this in future... I suppose once we get to 1.10 we can just have a `1\.1.+` pattern ?

PR #1318, opened 2025-07-21
chore(orchestrator): override backstage.json

batzionb:
enable publishing orchestrator even though upgrade of sources to 1.39.1 wasn't completed


davidfestal (changes requested):
You would also have to change the `repo-backstage-version` accordingly to `1.39.1` in the [`source.json`](https://github.com/redhat-developer/rhdh-plugin-export-overlays/blob/736ec63bdf44db4464843cb9bc450b04796fe3df/workspaces/orchestrator/source.json#L1) file.
This field is expected to reliably reflect the known backstage compatibility of the workspace sources.

davidfestal:
> > You would also have to change the `repo-backstage-version` accordingly to `1.39.1` in the [`source.json`](https://github.com/redhat-developer/rhdh-plugin-export-overlays/blob/736ec63bdf44db4464843cb9bc450b04796fe3df/workspaces/orchestrator/source.json#L1) file. This field is expected to reliably reflect the known backstage compatibility of the workspace sources.
> 
> done

That seems perfect.
Now you should be able to test that by triggering the publish action on the PR, by just adding the `/publish` comment.

davidfestal (approved):

PR #1324, opened 2025-07-23
Upgrade gitlab to `v6.12.1` like in the wrappers, with `1.39.1` as backstage compatible version.

davidfestal:


davidfestal:
Based on https://github.com/redhat-developer/rhdh/blob/release-1.7/dynamic-plugins/wrappers/immobiliarelabs-backstage-plugin-gitlab/package.json
cc @PatAKnight

PR #1361, opened 2025-08-05
FEAT RHIDP-7526: migrate lifecycle/support metadata to overlay repo for all supported plugins in downstream-plugins file

polasudo:
This pull request introduces a new script to automate the generation of Backstage metadata files for plugins and adds a generated metadata file for the 3scale backend plugin. The script applies specific logic to handle different plugin path structures and types, ensuring consistent and accurate metadata generation.

Automation and metadata generation:

* Added `generate_metadata.py`, a script that reads plugin paths from the `downstream-plugins` file, determines the correct output directory and plugin type, and generates a `metadata.yaml` file for each plugin with appropriate annotations and links. The script includes special logic to handle 'backstage' workspace plugins differently from others.

### DIFF
--- a/workspaces/3scale/metadata.yaml
+++ b/workspaces/3scale/metadata.yaml
@@ -1,0 +1,14 @@
+apiVersion: backstage.io/v1alpha1
+kind: Component
+metadata:
+  name: rhdh-plugin-3scale-backend
+  title: '@redhat/backstage-plugin-3scale-backend'
+  annotations:
+    janus-idp.io/lifecycle: production
+    janus-idp.io/support: supported
+    backstage.io/view-url: https://github.com/redhat-developer/rhdh-plugin-export-overlays/tree/main/workspaces/3scale
+    backstage.io/edit-url: https://github.com/redhat-developer/rhdh-plugin-export-overlays/edit/main/workspaces/3scale/metadata.yaml
+spec:
+  type: backstage-backend-plugin
+  owner: group:default/rhdh-maintainers
+  lifecycle: production
--- a/generate_metadata.py
+++ b/generate_metadata.py
@@ -1,0 +1,78 @@
+import os
+import re
+
+# --- Configuration ---
+downstream_plugins_file = 'downstream-plugins'
+overlay_repo_root = 'workspaces'
+repo_base_url = 'https://github.com/redhat-developer/rhdh-plugin-export-overlays'
+# --- End Configuration ---
+
+metadata_template = """apiVersion: backstage.io/v1alpha1
+kind: Component
+metadata:
+  name: rhdh-plugin-{plugin_name_sanitized}
+  title: '@redhat/backstage-plugin-{plugin_name_sanitized}'
+  annotations:
+    janus-idp.io/lifecycle: production
+    janus-idp.io/support: supported
+    backstage.io/view-url: {repo_base_url}/tree/main/workspaces/{final_path_for_url}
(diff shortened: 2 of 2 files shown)


workspaces/3scale/metadata.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: backstage.io/v1alpha1
+kind: Component
+metadata:
+  name: rhdh-plugin-3scale-backend
+  title: '@redhat/backstage-plugin-3scale-backend'
+  annotations:
+    janus-idp.io/lifecycle: production
```
nickboldt, line 7: hmm... not sure we want to have anything with janus-idp namespace.

Can we have files that contain just the lifecycle and support levels from the wrappers? eg., 

https://github.com/redhat-developer/rhdh/blob/main/dynamic-plugins/wrappers/backstage-community-plugin-acr/package.json#L63-L64

```
    "support:tech-preview",
    "lifecycle:active"
```

@davidfestal should this be implented as PATCHES for the package.json files, like you've done for pagerduty ? 

https://github.com/redhat-developer/rhdh-plugin-export-overlays/blob/release-1.7/workspaces/pagerduty-frontend/patches/1-set-version-in-package-json.patch

that way we can keep these changes in package.json of the published plugins.

Or could we add these as annotations on the plugins-list.yaml files?

thus in https://github.com/redhat-developer/rhdh-plugin-export-overlays/blob/release-1.7/workspaces/acr/plugins-list.yaml

```
plugins/acr:
  support: "tech-preview"
  lifecycle: "active"
```

Or maybe we want NEW json files for this?

workspaces/3scale/metadata.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: backstage.io/v1alpha1
+kind: Component
```
davidfestal, line 2: Why introducing a backstage catalog entity component syntax here, since marketplace catalog entities are already expected to be either stored in the overlay repository, or generated from data in the overlay.

workspaces/3scale/metadata.yaml
davidfestal: The problem with using the `downstream-plugin` is that it would only be applied to productized plugins, and not available for community plugins.

workspaces/3scale/metadata.yaml
polasudo: was not quite sure which one would suit so I used this, will rework it

workspaces/3scale/metadata.yaml
nickboldt: > The problem with using the downstream-plugin is that it would only be applied to productized plugins, and not available for community plugins.

If we rework the file as yaml we can include BOTH community and supported (TP/GA) in the same file. 

Could rename it to plugin-list.yaml; or have two files... one for community-plugins.yaml and one for downstream-plugins.yaml

polasudo: what would be the optimal solution ?

davidfestal: > I think that would imply that metadata/support information is associated with a specific release of RHDH, rather than plugins having independent lifecycle/support designations. Maybe that's what is desired?

I also guess that once we have the marketplace catalog entities (or at least to-be-completed skeletons of them) in the overlay, in the corresponding workspace folder, they would follow the changes done in the plugins themselves, and any change in those fields could be done at any moment in the main branch.

PR #1400, opened 2025-09-02
feat: Add workflow to auto-label workspace PRs

polasudo:
- Labels: workspace-update, workspace-addition, outside-workspaces, mandatory-workspace
- Detects required plugins from downstream-plugins file
- Runs daily at 6 AM UTC

Assisted-by: Cursor

### DIFF
--- a/.github/workflows/label-mandatory-workspace-prs.yaml
+++ b/.github/workflows/label-mandatory-workspace-prs.yaml
@@ -1,0 +1,243 @@
+name: Label Workspace PRs
+
+on:
+  schedule:
+    - cron: '0 6 * * *'  # Daily at 6:00 AM UTC
+  workflow_dispatch:  # Allow manual triggering
+
+concurrency:
+  group: ${{ github.workflow }}
+  cancel-in-progress: true
+
+jobs:
+  label-workspace-prs:
+    runs-on: ubuntu-latest
+    name: Label PRs based on Workspace Changes
+    
+    permissions:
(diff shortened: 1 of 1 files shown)


.github/workflows/label-mandatory-workspace-prs.yaml
```diff
@@ -0,0 +1,243 @@
...
+                
+                for (const file of prFiles.data) {
+                  const workspaceMatch = file.filename.match(/^workspaces\/([^\/]+)\/.*/);
+                  if (workspaceMatch) {
+                    const workspace = workspaceMatch[1];
+                    workspaceFiles.push({ file, workspace });
+                    
+                    // Check if this is a new workspace
+                    if (file.status === 'added') {
+                      addedWorkspaces.add(workspace);
```
davidfestal, line 147: Is this test sufficient to know that it is the addition of a full workspace that didn't exist before ?
What happens if the PR only add a new file to an already-existing workspace (like a patch file in the `/patches` sub-folder) ?

davidfestal (changes requested):
The logic that distinguishes between workspace addition and workspace modification might need some rework.

davidfestal (changes requested):
BTW, I think we should also add another label related to the branch the PR is based on.
So we should also do the following changes:
- Add the `mandatory-workspace` label only if the PR is based on the main branch
- Add a release-branch-patch label on workspace PRs based on a release branch

PR #1401, opened 2025-09-02
Add plugins/mcp-actions-backend to overlays

johnmcollier:
Adds the [mcp-actions-backend](https://github.com/backstage/backstage/tree/master/plugins/mcp-actions-backend) plugin from backstage/backstage to the plugin overlays, which will enable the MCP server backend plugin to be deployed on RHDH (provided the Backstage version is 1.40 or higher)

CC @davidfestal


johnmcollier:
Blocked on https://github.com/redhat-developer/rhdh-plugin-export-overlays/pull/1491

PR #1410, opened 2025-09-10
RHIDP-8834: Added Integration tests

rostalan:
Added rhdh deployment, plugin loading assertion.
Resolves [RHIDP-8834](https://issues.redhat.com/browse/RHIDP-8834)

### DIFF
--- a/.github/workflows/prepare-test-config.yaml
+++ b/.github/workflows/prepare-test-config.yaml
@@ -41,7 +41,10 @@ jobs:
           if_no_artifact_found: warn
           allow_forks: true
 
-      - name: Parse config and update OCI references
+      - name: Setup yq
+        uses: mikefarah/yq@v4
+
+      - name: Build dynamic-plugins.test.yaml
         run: |
           WORKSPACE_PATH="${{ inputs.workspace-path }}"
           PUBLISHED_EXPORTS=""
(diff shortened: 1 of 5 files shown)


.github/workflows/run-plugin-integration-tests.yaml
```diff
@@ -0,0 +1,139 @@
...
+          PLUGINS=$(grep -Eo '!([^[:space:]]+)' ./artifacts/dynamic-plugins.test.yaml | sed 's/^!//' | sort -u)
+          if [ -z "$PLUGINS" ]; then
+            echo "No plugins found in dynamic-plugins.test.yaml"; exit 1
+          fi
+          LOGS=$(docker logs rhdh || true)
+          # Check each plugin appears as loaded in logs
+          echo "===== Checking logs for plugin loaded messages"
+          for plugin in $PLUGINS; do
+            echo "Asserting plugin loaded: $plugin"
+            if echo "$LOGS" | grep -qi "loaded dynamic .* plugin" && echo "$LOGS" | grep -q "/dynamic-plugins-root/${plugin}"; then
```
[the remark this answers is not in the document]
rostalan: yes, it should. https://github.com/rostalan/rhdh-plugin-export-overlays/actions/runs/17610591280/job/50032417785#step:9:36

.github/workflows/update-test-config.yaml
```diff
@@ -0,0 +1,93 @@
...
+              echo "Export did not match expected format, skipping: $export"
+            fi
+          done
+
+      - name: Prepare artifacts for integration tests
+        run: |
+          mkdir -p ./integration-artifacts
+          
+          # Copy dynamic-plugins test config
+          if [ -f "${{ format('{0}/tests/dynamic-plugins.test.yaml', inputs.workspace-path) }}" ]; then
```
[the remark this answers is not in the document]
rostalan: yes

.github/workflows/prepare-test-config.yaml
```diff
@@ -0,0 +1,130 @@
...
+          fi
+
+          # copy optional workspace-level test configs into artefact
+          [ -f "$WORKSPACE_PATH/tests/app-config.test.yaml" ] && cp "$WORKSPACE_PATH/tests/app-config.test.yaml" "$OUT_DIR/"
+          [ -f "$WORKSPACE_PATH/tests/test.env" ] && cp "$WORKSPACE_PATH/tests/test.env" "$OUT_DIR/"
+
+          echo "Contents of $OUT_DIR after copy:" && ls -la "$OUT_DIR"
+
+          # Install yq for YAML processing
+          sudo curl -sL https://github.com/mikefarah/yq/releases/latest/download/yq_linux_amd64 -o /usr/local/bin/yq
```
davidfestal, line 80: It would be better if we could void installing this this way.

.github/workflows/pr-actions.yaml
```diff
@@ -183,6 +183,40 @@ jobs:
...
+          {"workspace":"${WORKSPACE}","overlayBranch":"${OVERLAY_BRANCH}","overlayRepo":"${OVERLAY_REPO}"}
+          EOF
+      - name: Upload published-exports artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: published-exports
+          path: published-exports
+          if-no-files-found: error
+          retention-days: 7
+
```
davidfestal, line 219: Where is the addition of the triggering of tests through he `/test` comment ?

.github/workflows/prepare-test-config.yaml
```diff
@@ -0,0 +1,130 @@
...
+          # For each published export, extract plugin name and read its metadata
+          # Expected export format: ghcr.io/<repo_path>/<plugin_name>:<tag>
+          for export in $PUBLISHED_EXPORTS; do
+            echo "Processing export: $export"
+            if [[ "$export" =~ ^ghcr\.io/(.+):([^[:space:]]+)$ ]]; then
+              IMAGE_PATH_AND_PLUGIN="${BASH_REMATCH[1]}"   # <repo_path>/<plugin_name>
+              NEW_TAG="${BASH_REMATCH[2]}"                 # <tag>
+              PLUGIN_NAME="${IMAGE_PATH_AND_PLUGIN##*/}"
+
+              METADATA_FILE="$WORKSPACE_PATH/metadata/${PLUGIN_NAME}.yaml"
```
davidfestal, line 95: I'm not sure it would be guaranteed that the name of the yaml file in the `metadata` folder would always be exactly the same as the one extracted from the image name.
But re-creating the container image name from the package name contained in the metadata yaml and using that to match metadata yaml files with OCI references seems more robust.

davidfestal: (because we know the logic of the oci reference name from the package name, so that we can reproduce it here.)

.github/workflows/pr-actions.yaml
rostalan: funnily enough, it is not explicitely needed here or anywhere else, parse and prepare jobs run on any command and when they are executed on running `/test`, the success workflow state triggers the integration tests. So it would seem it just needs to be added as a valid command.

PR #1627, opened 2025-11-21
fix(tests): changed the expected format of the config files

rostalan:
- https://issues.redhat.com/browse/RHIDP-9524
- updated the integration tests to reflect changes in `workspaces/<workspace_name>/metadata/<plugin_name>.yaml` file format.
- added env var file (`workspaces/<workspace_name>/tests/test.env`) checks, it passes if the file is found and contains all needed vars, skips when no file is found and required and fails if any of the required vars are missing.
- updated `README.md`


.github/workflows/workspace-tests.yaml
```diff
@@ -155,24 +169,62 @@ jobs:
...
+              # If no ENVs in config, continue regardless of env file existence
+
               STRIPPED=$(echo "$PACKAGE_NAME" | sed 's|^@||; s|/|-|')
               echo "- package: \"oci://ghcr.io/${IMAGE_PATH_AND_PLUGIN}:${NEW_TAG}!${STRIPPED}\"" >> "$OUT_FILE"
               echo "  disabled: false" >> "$OUT_FILE"
               echo "  pluginConfig:" >> "$OUT_FILE"
-              # Append the configs.default block indented by 4 spaces under pluginConfig
-              yq -o=yaml '.configs.default' "$METADATA_FILE" | sed 's/^/    /' >> "$OUT_FILE"
+              # Append the first appConfigExamples content block indented by 4 spaces under pluginConfig
+              echo "$CONFIG_CONTENT" | sed 's/^/    /' >> "$OUT_FILE"
```
davidfestal: Seems to me that using jq & yq, which afaik are included by default in the default ubuntu runner, would be more robust ?

davidfestal (changes requested):
Do you really need to install `yq` with:

```yaml
      - name: Setup yq
        uses: mikefarah/yq@v4
```

knowing that it is included by default in the ubuntu runner (afaict) ?

.github/workflows/workspace-tests.yaml
davidfestal: `paths` is marked with an error (`unexpected value`) in my yaml editor, according to the schema.
I looked into the schema, and `paths` doesn't seem available for `workflow_run` events

.github/workflows/run-plugin-integration-tests.yaml
```diff
@@ -2,10 +2,6 @@ name: Run Plugin Integration Tests
 
 on:
   workflow_call:
```
davidfestal: What is the reason of removing the `workspace-path` input here ?

davidfestal (approved):

PR #1973, opened 2026-02-25
chore(tests) updated the rhdh and plugin error detection

rostalan:
- added error checking and backstage startup checking
- it will require config to be added for all plugins to all workspaces


rostalan:
closed as superceeded by https://github.com/redhat-developer/rhdh-plugin-export-overlays/pull/2231

PR #2026, opened 2026-03-11
chore: update orchestrator repo-ref

lokanandaprabhu:
update orchestrator repo-ref 06ff6dfa5bf9fafe1043b2572598fe40f75952d6


davidfestal:
@lokanandaprabhu Does this PR still make sense after the merged of the following automatically-created PR: https://github.com/redhat-developer/rhdh-plugin-export-overlays/pull/2162 ?

BTW, changing the commit reference by hand is not a recommended practice, and should be avoided as much as possible. There are automatic workflow, that can also be triggered manually if necessary, to upgrade the commit to the latest commit published on NPM.
Setting the commit by hand to another commit can very easily bring conflicts, break the automatic discovery workflow process and finally require time-consuming work to manually look into the commit history.

lokanandaprabhu:
@davidfestal this is not required. 

>BTW, changing the commit reference by hand is not a recommended practice, and should be avoided as much as possible. There are automatic workflow, that can also be triggered manually if necessary, to upgrade the commit to the latest commit published on NPM.
Setting the commit by hand to another commit can very easily bring conflicts, break the automatic discovery workflow process and finally require time-consuming work to manually look into the commit history.

Noted

PR #2069, opened 2026-03-30
chore(e2e): migrate techdocs plugins e2e tests

PatAKnight:
## Description

Moves the techdocs plugins e2e tests over from the RHDH repo to the overlays repo.

Updates the scaffolder backend module azure plugin to point to the OCI artifact since we no longer include it as a dynamic plugin wrapper within RHDH.

## Which issue(s) does this PR fix

- Fixes [RHIDP-11387](https://redhat.atlassian.net/browse/RHIDP-11387)

### DIFF
--- a/workspaces/backstage/e2e-tests/tests/config/techdocs/dynamic-plugins.yaml
+++ b/workspaces/backstage/e2e-tests/tests/config/techdocs/dynamic-plugins.yaml
@@ -1,0 +1,7 @@
+plugins:
+  - package: ./dynamic-plugins/dist/backstage-plugin-techdocs
+    disabled: false
+  - package: ./dynamic-plugins/dist/backstage-plugin-techdocs-backend-dynamic
+    disabled: false
+  - package: ./dynamic-plugins/dist/backstage-plugin-techdocs-module-addons-contrib
+    disabled: false
--- a/workspaces/backstage/e2e-tests/tests/config/techdocs/dynamic-plugins.yaml
+++ b/workspaces/backstage/e2e-tests/tests/config/techdocs/dynamic-plugins.yaml
@@ -1,7 +1,7 @@
 plugins:
   - package: ./dynamic-plugins/dist/backstage-plugin-techdocs
-    disabled: false
+    disabled: true
   - package: ./dynamic-plugins/dist/backstage-plugin-techdocs-backend-dynamic
-    disabled: false
+    disabled: true
   - package: ./dynamic-plugins/dist/backstage-plugin-techdocs-module-addons-contrib
-    disabled: false
+    disabled: true
--- a/workspaces/backstage/e2e-tests/playwright.config.ts
+++ b/workspaces/backstage/e2e-tests/playwright.config.ts
@@ -1,7 +1,4 @@
 import { defineConfig } from "@red-hat-developer-hub/e2e-test-utils/playwright-config";
-import dotenv from "dotenv";
-
-dotenv.config({ path: `${import.meta.dirname}/.env` });
 
 /**
  * Backstage workspace e2e test configuration.
@@ -10,19 +7,19 @@ dotenv.config({ path: `${import.meta.dirname}/.env` });
 export default defineConfig({
   projects: [
     {
-      name: 'backstage-github-org-discovery',
+      name: "backstage-github-org-discovery",
       testMatch: /tests\/specs\/github-org-discovery\.spec\.ts/,
     },
     {
-      name: 'backstage-github-discovery',
+      name: "backstage-github-discovery",
       testMatch: /tests\/specs\/github-discovery\.spec\.ts/,
     },
     {
-      name: 'backstage-gitlab-discovery',
+      name: "backstage-gitlab-discovery",
       testMatch: /tests\/specs\/gitlab-discovery\.spec\.ts/,
     },
     {
-      name: 'backstage-github-events',
+      name: "backstage-github-events",
       testMatch: /tests\/specs\/github-events-module\.spec\.ts/,
     },
     {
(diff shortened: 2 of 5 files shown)


workspaces/backstage/e2e-tests/tests/config/techdocs/dynamic-plugins.yaml
```diff
@@ -0,0 +1,7 @@
+plugins:
+  - package: ./dynamic-plugins/dist/backstage-plugin-techdocs
+    disabled: true
+  - package: ./dynamic-plugins/dist/backstage-plugin-techdocs-backend-dynamic
+    disabled: true
```
subhashkhileri, line 5: @PatAKnight are these plugins supposed to be disabled? I’m wondering how the test passed if they were disabled

PatAKnight: @subhashkhileri for some reason I had to disable the wrappers to get the tests to pass. It seems like the tests were trying to load both the techdocs wrappers and oci artifacts at the same time. I kept getting the error `Plugin 'techdocs' startup failed; caused by Error: ExtensionPoint with ID 'techdocs.builds' is already registered`

subhashkhileri: got it. I can see that these plugins are enabled by default as wrappers, and when OCI is enabled, it leads to the error above. A temporary solution is to disable those wrappers, similar to this example: https://github.com/redhat-developer/rhdh-plugin-export-overlays/blob/main/workspaces/adoption-insights/e2e-tests/tests/specs/adoption-insights.spec.ts#L31

This is just an interim fix until the wrappers are removed: https://github.com/redhat-developer/rhdh-e2e-test-utils/pull/62

With this, the PR checks will force-disable the wrappers and use OCI, while nightly runs will continue testing the wrappers.

Also, we can check the stdio in the Playwright report to see which configs were used. From what I see, OCI refs were disabled, meaning it was still using the wrapper during the PR check hence test passed, whereas it should have been using OCI.
```
┌─ Dynamic Plugins ───────────────────────────────────────────────────
includes:
  - dynamic-plugins.default.yaml
plugins:
  - package: ./dynamic-plugins/dist/backstage-community-plugin-catalog-backend-module-keycloak-dynamic
    disabled: false
  - package: oci://ghcr.io/redhat-developer/rhdh-plugin-export-overlays/backstage-plugin-techdocs:pr_2069__1.17.2!backstage-plugin-techdocs
    disabled: true
    pluginConfig:
      dynamicPlugins:
        frontend:
          backstage.plugin-techdocs:
            routeBindings:
              targets:
                - importName: techdocsPlugin
              bindings:
                - bindTarget: catalogPlugin.externalRoutes
                  bindMap:
                    viewTechDoc: techdocsPlugin.routes.docRoot
                - bindTarget: scaffolderPlugin.externalRoutes
                  bindMap:
                    viewTechDoc: techdocsPlugin.routes.docRoot
            dynamicRoutes:
              - path: /docs
                importName: TechDocsIndexPage
                menuItem:
                  icon: docs
                  text: Docs
                  textKey: menuItem.docs
              - path: /docs/:namespace/:kind/:name/*
                importName: TechDocsReaderPage
            mountPoints:
              - mountPoint: entity.page.docs/cards
                importName: EntityTechdocsContent
                config:
                  layout:
                    gridColumn: 1 / -1
                  if:
                    allOf:
                      - isTechDocsAvailable
              - mountPoint: search.page.results
                importName: TechDocsSearchResultListItem
              - mountPoint: search.page.filters
                importName: TechdocsSearchFilter
              - mountPoint: search.page.types
                importName: techdocsSearchType
                config:
                  props:
                    name: Documentation
                    icon: docs
  - package: oci://ghcr.io/redhat-developer/rhdh-plugin-export-overlays/backstage-plugin-techdocs-backend:pr_2069__2.1.6!backstage-plugin-techdocs-backend
    disabled: true
    pluginConfig:
      techdocs:
        builder: local
        generator:
          runIn: local
        publisher:
          type: local
  - package: oci://ghcr.io/redhat-developer/rhdh-plugin-export-overlays/backstage-plugin-techdocs-module-addons-contrib:pr_2069__1.1.34!backstage-plugin-techdocs-module-addons-contrib
    disabled: true
    pluginConfig:
      dynamicPlugins:
        frontend:
          backstage.plugin-techdocs-module-addons-contrib:
            techdocsAddons:
              - importName: ReportIssue

└──────────────────────────────────────────────────────────────────────────────```

PatAKnight: I didn't know about the `disableWrappers` option. I can swap to using that over this dedicated `dynamic-plugings.yaml`.

subhashkhileri: i guess having a dedicated `dynamic-plugings.yaml` with just package refs and disabled : false should be okay just for this workspace because this backstage workspace have more than just tech docs plugins and when no dedicated dynamic-plugings.yaml is provided it will enable all the plugins from this workspace which i guess you don't want just to test tech-docs.

https://redhat-developer.github.io/rhdh-e2e-test-utils/overlay/test-structure/configuration-files.html#when-file-exists

PatAKnight: I am fine with either approach. Having all of the plugins end up being enabled does lead to a lot of noise in the logs. So, I am a bit partial to the `dynamic-plugins.yaml` file but either is fine with me.

lholmquist (approved):

PR #2231, opened 2026-04-09
feat(poc): replace Docker-based smoke tests with native Node.js harness

rostalan:
Boot a minimal Backstage backend directly on the runner using `createBackend()` + `dynamicPluginsFeatureLoader`, probe `/api/<pluginId>` routes, and report results as structured JSON. Includes core bundled plugins (catalog, auth, permission, scaffolder, events, search, proxy) so dynamic plugins resolve their dependencies correctly.

Frontend plugins are also validated via two layers:
- **Static bundle checks** — verifies `dist-scalprum/` exists and contains JavaScript files after OCI download
- **Loaded-plugin probe** — an inline backend plugin queries `dynamicPluginsServiceRef` to confirm frontend plugins were registered without errors

Results are merged into a single report with per-plugin status (`pass`, `fail-bundle`, `fail-load`, `warn`, `skip`) and the process exits non-zero on any failure.

Made-with: Cursor


smoke-tests/smoke-test.mjs
```diff
@@ -0,0 +1,681 @@
...
+}
+
+const FRONTEND_ROLES = new Set(["frontend-plugin", "frontend-plugin-module"]);
+
+function isFrontendRole(role) {
+  return FRONTEND_ROLES.has(role);
+}
+
+// ---------------------------------------------------------------------------
+// OCI Download  (mirrors install-dynamic-plugins.py §663-715)
```
gustavolira: The comment itself acknowledges this mirrors `install-dynamic-plugins.py`. Should we just call the Python script directly as a pre-step instead of reimplementing it in JS?

The overlay repo currently gets the script from the RHDH container image — by removing Docker, we lose access to it. But we could copy the script here (with a version marker) as a short-term solution. That would:
- Eliminate ~60 lines of JS OCI code
- Inherit all security protections for free (zip bomb, symlink traversal, integrity checks)
- Leverage the 130KB+ test suite that already covers this logic

Longer-term options: publish as a pip package or container tool so any repo can consume it.

For context, there's also a parallel POC in RHDH core (redhat-developer/rhdh#4523) that extends the same Python script with parallel downloads (`ThreadPoolExecutor`) achieving ~34% speedup. Reusing it here would get that performance benefit too.

smoke-tests/smoke-test.mjs
```diff
@@ -0,0 +1,681 @@
...
+  const manifest = JSON.parse(
+    readFileSync(join(localDir, "manifest.json"), "utf8"),
+  );
+  const [, filename] = manifest.layers[0].digest.split(":");
+  return join(localDir, filename);
+}
+
+function extractPlugin(tarFile, pluginPath, dest) {
+  mkdirSync(join(dest, pluginPath), { recursive: true });
+  execSync(`tar xf "${tarFile}" -C "${dest}" "${pluginPath}/"`, {
```
gustavolira: This extracts tar contents with a bare `execSync('tar xf ...')` — no validation of archive members before extraction. The Python script in RHDH core (`install-dynamic-plugins.py`) checks for:

- **Zip bomb**: `member.size > MAX_ENTRY_SIZE` (20MB default per entry)
- **Symlink traversal**: `os.path.realpath()` validation against the destination directory
- **Hardlink traversal**: same check
- **Device files/FIFOs**: rejected entirely
- **Safe tar filter**: `tar.extractall(..., filter='tar')` 

Even in CI, a crafted OCI layer could exploit path traversal via symlinks. If we keep the JS implementation, all of these checks need to be ported.

smoke-tests/smoke-test.mjs
```diff
@@ -0,0 +1,681 @@
...
+// Backend boot
+// ---------------------------------------------------------------------------
+
+async function bootBackend(configPaths) {
+  process.argv = ["node", "smoke-test.mjs"];
+  for (const p of configPaths) {
+    if (existsSync(p)) process.argv.push("--config", p);
+  }
+
+  const { createBackend } = await import("@backstage/backend-defaults");
```
gustavolira: `createBackend()` is the production API. Backstage provides `startTestBackend()` from `@backstage/backend-test-utils` specifically for this use case — it handles lifecycle management, automatic cleanup, built-in in-memory SQLite, and exposes mock services. It would also reduce the manual plugin registration below (~20 lines of `backend.add()` calls).

redhat-developer/rhdh#4523 uses `startTestBackend()` for the same plugin loadability validation and it works well.

smoke-tests/smoke-test.mjs
```diff
@@ -0,0 +1,681 @@
...
+    results.push({ pkgName, role, pluginPath, status: "pass" });
+  }
+  return results;
+}
+
+// ---------------------------------------------------------------------------
+// Config generation
+// ---------------------------------------------------------------------------
+
+function deepMerge(src, dst) {
```
gustavolira: This reimplements config deep-merge that already exists in `install-dynamic-plugins.py` (the `merge()` function). The Python version also detects key conflicts and raises errors — this one silently overwrites. If we use the Python script as a pre-step, this function becomes unnecessary since the script already generates `app-config.dynamic-plugins.yaml` with all plugin configs merged.

smoke-tests/smoke-test.mjs
```diff
@@ -0,0 +1,681 @@
...
+  backend.add(
+    createServiceFactory({
+      service: dynamicPluginsFrontendServiceRef,
+      deps: {},
+      factory: () => ({ setResolverProvider() {} }),
+    }),
+  );
+
+  backend.add(
+    createBackendPlugin({
```
gustavolira: This probe plugin concept is great — using `dynamicPluginsServiceRef` to verify frontend plugins actually loaded at runtime is more thorough than filesystem-only checks. Worth keeping regardless of the OCI download approach. The sanity check POC in RHDH core (redhat-developer/rhdh#4523) currently only validates bundles via filesystem (`dist-scalprum/plugin-manifest.json`); this runtime probe approach would complement it nicely.

smoke-tests/smoke-test.mjs
```diff
@@ -0,0 +1,681 @@
...
+}
+
+function extractPlugin(tarFile, pluginPath, dest) {
+  mkdirSync(join(dest, pluginPath), { recursive: true });
+  execSync(`tar xf "${tarFile}" -C "${dest}" "${pluginPath}/"`, {
+    stdio: "pipe",
+  });
+}
+
+async function downloadPlugins(plugins, dest) {
```
gustavolira: Downloads are sequential here. The `install-dynamic-plugins-fast.py` in redhat-developer/rhdh#4523 uses `ThreadPoolExecutor` for parallel OCI downloads with a shared image cache (one download per unique image, not per plugin), achieving ~34% speedup. Another reason to reuse the existing script rather than maintaining a separate implementation.

smoke-tests/smoke-test.mjs
rostalan: for now, I added a step that downloads install-dynamic-plugins script and its requirements and uses it directly. Not sure if the would benefit from the parallel downloads, since it only runs on a single workspace, so only a couple OCIs at a time.

smoke-tests/smoke-test.mjs
rostalan: replaced by python script.

smoke-tests/smoke-test.mjs
rostalan: This was meant to keep the environment somewhat real, but I guess it makes sense to have this as a simple "plugin loads" check rather than anything more complicated.

smoke-tests/smoke-test.mjs
rostalan: replaced with python script

smoke-tests/smoke-test.mjs
rostalan: replaced with the basic python script, same reason as above.

.github/workflows/run-workspace-smoke-tests.yaml
```diff
@@ -36,139 +36,105 @@ jobs:
           name: smoke-test-artifacts
           path: ./artifacts
 
-      - name: Log in to GitHub Container Registry
-        uses: docker/login-action@4907a6ddec9925e35a0a9e82d7399ccc52663121 # v4.1.0
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
```
jonkoops: This should be pinned to a hash and match the other workflows:
```suggestion
        uses: actions/setup-node@53b83947a5a98c8d113130e565377fae1a50d02f # v6.3.0
```

.github/workflows/run-workspace-smoke-tests.yaml
```diff
@@ -36,139 +36,105 @@ jobs:
...
         with:
-          registry: ghcr.io
-          username: ${{ github.actor }}
-          password: ${{ secrets.GITHUB_TOKEN }}
+          node-version: "20"
```
jonkoops: There are two LTS releases after this one that are currently active, let's not target older Node.js versions if we can avoid it.
```suggestion
          node-version: "24"
```

.github/workflows/run-workspace-smoke-tests.yaml
```diff
@@ -36,139 +36,105 @@ jobs:
...
-            if ! docker ps | grep -q rhdh; then
-              echo "Container stopped unexpectedly."
-              exit 1
-            fi
-            sleep 10
-          done
-          echo "RHDH did not become ready in time."
-          exit 1
+          curl -fsSL \
+            "https://raw.githubusercontent.com/redhat-developer/rhdh/3efb9cc140ff/scripts/install-dynamic-plugins/install-dynamic-plugins.py" \
```
jonkoops, line 75: Is this supposed to be pinned to a specific commit?

smoke-tests/package.json
```diff
@@ -0,0 +1,26 @@
+{
+  "name": "rhdh-smoke-test",
+  "version": "0.1.0",
+  "private": true,
+  "type": "module",
+  "description": "Lightweight smoke test harness for RHDH dynamic plugins — no Docker required",
+  "scripts": {
+    "test": "node smoke-test.mjs",
```
jonkoops: Looks like this script is not triggered in the CI, but rather executed manually, let's use `npm test` for consistency.

smoke-tests/package.json
```diff
@@ -0,0 +1,26 @@
...
+  "dependencies": {
+    "@backstage/backend-dynamic-feature-service": "0.7.9",
+    "@backstage/backend-plugin-api": "1.7.0",
+    "@backstage/backend-test-utils": "1.4.0",
+    "@backstage/cli-node": "0.2.18",
+    "@backstage/config": "1.3.6",
+    "@backstage/config-loader": "1.10.8",
+    "@backstage/errors": "1.2.7",
+    "@backstage/types": "1.2.2",
+    "js-yaml": "4.1.0"
```
jonkoops, line 21: Let's not rely on pinned versions here, but use caret ranges `^` and a lockfile.

smoke-tests/smoke-test.ts
jonkoops, line 1: There is no need to use `.mjs`, you can use `.js` and Node will be able to detect it, also because the type is already set in the package.

jonkoops, line 1: Node.js can run TypeScript these days, so using `.ts`  would also work, and make your code type-safe.

smoke-tests/smoke-test.mjs
```diff
@@ -0,0 +1,597 @@
...
+import { fileURLToPath } from "node:url";
+import yaml from "js-yaml";
+
+const __dirname = dirname(fileURLToPath(import.meta.url));
+
+// ---------------------------------------------------------------------------
+// CLI
+// ---------------------------------------------------------------------------
+
+function parseArgs(argv) {
```
jonkoops: Node.js has [built-in support](https://nodejs.org/api/util.html#utilparseargsconfig) for parsing CLI arguments, there is no need to implement this yourself.

smoke-tests/smoke-test.mjs
```diff
@@ -0,0 +1,597 @@
...
+  if (!args.pluginsYaml) {
+    console.error(
+      "Usage: node smoke-test.mjs --plugins-yaml <path> [--config <path>...] [--env-file <path>] [--skip-download]",
+    );
+    process.exit(1);
+  }
+  return args;
+}
+
+function loadEnvFile(filePath) {
```
jonkoops: Node.js has [built-in support](https://nodejs.org/api/environment_variables.html#dotenv) for parsing `.env` files, so there is no need to implement this yourself.

smoke-tests/smoke-test.mjs
```diff
@@ -0,0 +1,616 @@
...
+
+function isFrontendRole(role) {
+  return FRONTEND_ROLES.has(role);
+}
+
+// ---------------------------------------------------------------------------
+// Config helpers
+// ---------------------------------------------------------------------------
+
+function deepMerge(src, dst) {
```
jonkoops: The custom `deepMerge` and `loadConfigs` can be removed entirely — Backstage's config loader already handles deep merging of multiple `--config` files. Just push the config paths onto `process.argv` before calling `startTestBackend()` and let the default `rootConfig` service pick them up. This is the same approach the old Docker-based smoke test used (chaining `--config` flags on the `node packages/backend` command).

This also removes the need for `mockServices.rootConfig.factory({ data })`.

.github/workflows/run-workspace-smoke-tests.yaml
rostalan: this will get replaced with the ts version after https://github.com/redhat-developer/rhdh/pull/4574 gets merged anyway...

jonkoops (approved):

davidfestal (commented):
TBF I'm not convinced of this change.
Wouldn't it be better at least to wait for when the NFS becomes the default for fronted plugins, and we would be able to directly start a dynamic-plugins-enabled Backstage (not even RHDH) of the target backstage version ?

PR #2247, opened 2026-04-10
Register matomo analytics api

mjholder:
This PR creates the overlay directory needed for Matomo Analytics. Right now the plugin loads during the startup of RHDH but it never registers the plugins API so it doesn't attempt to send usage data to our Matomo server.


mjholder:
I'm going to focus on updating the upstream plugin to work with the wrapper

davidfestal:
Closing since he content of this PR has been included in PR https://github.com/redhat-developer/rhdh-plugin-export-overlays/pull/2427

PR #2251, opened 2026-04-14
chore: automate workspace labels for PRs

PatAKnight:
## Description

Adds automated PR labeling based on modified workspaces. This change introduces a workflow that monitors file changes and applies relevant workspace labels automatically. This ensures better visibility into which parts of the monorepo are being affected without requiring manual tagging by the author. 

This automation does not dynamically create labels in the repository settings. Will need to be some manual effort but if everyone agrees with these changes, I can go in and begin the process.


gashcrumb (approved):

.github/labeler.yml
```diff
@@ -0,0 +1,197 @@
...
+  - 'Workspace\\s*rbac'
+
+workspace/roadie-backstage-plugins:
+  - 'Workspace\\s*roadie-backstage-plugins'
+
+workspace/scaffolder-backend-module-kubernetes:
+  - 'Workspace\\s*scaffolder-backend-module-kubernetes'
+
+workspace/scaffolder-backend-module-regex:  
+- 'Workspace\\s*scaffolder-backend-module-regex'
```
gashcrumb: missing a tab here looks like

.github/workflows/add-workspace-label.yaml
```diff
@@ -0,0 +1,47 @@
...
+        with:
+          egress-policy: audit
+
+      - uses: github/issue-labeler@c1b0f9f52a63158c4adc09425e858e87b32e9685 #v3.4
+        with:
+          configuration-path: .github/labeler.yml
+          enable-versioned-regex: 0
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+
+  label-prs:
```
gashcrumb, line 33: Should this be gated to only run on pull requests, i.e.:

```
if: github.event_name == 'pull_request_target'
```

gashcrumb (changes requested):
one line's off in the `labeler.yml` file, the other issue is likely minor.

.github/labeler.yml
PatAKnight: Fixed

.github/workflows/add-workspace-label.yaml
PatAKnight: Fixed and also opened a PR to fix my mistake in the rhdh-plugins repo: https://github.com/redhat-developer/rhdh-plugins/pull/2892

PR #2309, opened 2026-04-24
fix(e2e): add plugin config for github tests

jrichter1:
fix deployment issues by only enabling the plugins that are being tested


teknaS47 (approved):
/lgtm

PR #2331, opened 2026-04-28
x2a: add default appConfigExamples

mareklibra:


mareklibra:
@gashcrumb @davidfestal Can you please have a look at this PR?

davidfestal (changes requested):
We don't need the `app-config.dynamic.yaml` files anymore. Their content should already be in the `metadata/*.yaml` files

workspaces/x2a/plugins/x2a-dcr/app-config.dynamic.yaml
```diff
@@ -0,0 +1,6 @@
+dynamicPlugins:
```
dzemanov: You should not need `app-config.dynamic.yaml` files as we do have https://redhat.atlassian.net/browse/RHDHPLAN-1125 to remove those. CC @gashcrumb @nickboldt these files are not used anywhere?

PR #2341, opened 2026-04-28
chore: update "catalog integration" extensions metadata titles and descriptions

kim-tsao:
Fixes: https://redhat.atlassian.net/browse/RHDHPLAN-365 and https://redhat.atlassian.net/browse/RHIDP-13267

* Update the metadata yaml titles to match the titles in the workspace metadata
* Corrected the plugin descriptions
* Corrected the docs URL (they were broken)

### DIFF
--- a/catalog-entities/extensions/plugins/github-catalog-integration.yaml
+++ b/catalog-entities/extensions/plugins/github-catalog-integration.yaml
@@ -4,7 +4,7 @@ kind: Plugin
 metadata:
   name: backstage-plugin-catalog-backend-module-github
   namespace: rhdh
-  title: Catalog Backend Module GitHub
+  title: GitHub Catalog Entity Discovery
   description: |
     Provides a GitHub integration for the Software Catalog. Searches your
     GitHub organization and registers catalog files matching the configured path.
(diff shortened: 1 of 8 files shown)


catalog-entities/extensions/plugins/github-catalog-integration.yaml
```diff
@@ -4,11 +4,10 @@ kind: Plugin
 metadata:
   name: backstage-plugin-catalog-backend-module-github
   namespace: rhdh
-  title: GitHub Software Catalog Integration
+  title: Catalog Backend Module GitHub
```
JessicaJHee:
```suggestion
  title: GitHub Catalog Entity Discovery
```
A small suggestion, Imho I think we should name these plugin entries to something more user friendly and focused on the functionality rather than aligning with the internal module naming scheme, wdyt?

JessicaJHee (commented):
Just a small comment on the title of all of these plugins, otherwise LGTM!

catalog-entities/extensions/plugins/github-catalog-integration.yaml
kim-tsao: How about `Catalog Backend Module GitHub Discovery` ?

JessicaJHee (approved):
LGTM 👍

PR #2376, opened 2026-05-01
refactor: merge aws-ecs and aws-codebuild into backstage-plugins-for-aws

hopehadfield:
Combines the `aws-ecs` and `aws-codebuild` overlay workspaces into a single directory, `workspaces/backstage-plugins-for-aws`, because both use the same upstream repo ([awslabs/backstage-plugins-for-aws](https://github.com/awslabs/backstage-plugins-for-aws)). One `source.json` / patch set / export target avoids duplicated pins and drift between the two.

### Changes

- Add `workspaces/backstage-plugins-for-aws` with `merged plugins-list.yaml` (ECS + CodeBuild), shared `patches/1-avoid-double-wildcards.patch`, `source.json`, `backstage.json`, `smoke-tests/test.env`, and metadata for all four published packages.
- Set `spec.partOf` on those Package entities to `backstage-plugins-for-aws`.
- Update CODEOWNERS to `/workspaces/backstage-plugins-for-aws` (replacing `aws-codebuild`).
- Remove `workspaces/aws-ecs` and `workspaces/aws-codebuild`.

cc @davidfestal


hopehadfield:
Once this is merged, we can add the `enableScripts: false` patch (I tested it locally and it worked), which should resolved the [failed export job](https://github.com/redhat-developer/rhdh-plugin-export-overlays/actions/runs/25188931601/job/73853767074) @davidfestal

davidfestal (approved):

davidfestal:
I guess we would then update PR https://github.com/redhat-developer/rhdh-plugin-export-overlays/pull/1873 to include `aws-codepipeline` plugins in the same `backstage-plugins-for-aws` overlay workspace.

durandom (approved):
I'm fine with that :)

PR #2378, opened 2026-05-04
fix(e2e): fix techdocs failing tests

PatAKnight:
Fix the failing techdocs failing tests. This PR updates the dedicated `dynamic-plugins.yaml` file and adds `disableWrappers`.


subhashkhileri (approved):

gashcrumb:
Hey @PatAKnight think this config needs to be added to the test's `app-config-rhdh.yaml`:

```
      dynamicPlugins:
        frontend:
          backstage.plugin-techdocs-module-addons-contrib:
            techdocsAddons:
              - importName: ReportIssue
```

We used to provide this in the `dynamic-plugins.default.yaml` file like [this](https://github.com/redhat-developer/rhdh/blob/release-1.7/dynamic-plugins.default.yaml#L770-L777) but likely this is no longer the case.

gashcrumb:
hmm, nope I see it in the config in the build output, will keep looking :-)

gashcrumb:
I wonder if the `index.ts` overlay file is getting picked up given these lines from the playwright trace:

```
Plugin backstage.plugin-techdocs is not configured properly: PluginRoot.TechdocsSearchFilter not found, ignoring mountPoint: "search.page.filters"
Plugin backstage.plugin-techdocs is not configured properly: PluginRoot.techdocsSearchType not found, ignoring mountPoint: "search.page.types"
```

As these are being exported [here](https://github.com/redhat-developer/rhdh-plugin-export-overlays/pull/2378/changes#diff-7bf1aa6cf7609fadf91a94cac73987c426bd06eed9da7226d72211f0301b7a66R52).  If not then it could be the plugin's real techdocs page export is being used and not the wrapped one.

I'll have a look through the `publish` output and see if there's anything obvious there.

gashcrumb:
Hmm, [this](https://github.com/redhat-developer/rhdh-plugin-export-overlays/actions/runs/25702883218/job/75466781713#step:17:1386) seems okay and as expected.  I'll see if I can give this step a run locally so I can see what the dist-dynamic folder looks like after the plugin export.

gashcrumb:
Downloaded the techdocs OCI artifact from this PR and ran ` rg -n "TechdocsSearchFilter" .` in the root, and I don't see any mention of this string which is unexpected, this should be present in both the federated module `dist` and scalprum build under `dist-scalprum` as an export.  Spending some time now confirming what happens after the overlay files are applied, maybe there needs to be a build step after this.

gashcrumb:
@PatAKnight Okay, think I have it.  Move the source overlay files into a `src` directory.  Right now the overlay files are being copied to the root of the plugin directory instead of overlaying `src/index.ts` as we would want.

dzemanov:
> @PatAKnight Okay, think I have it. Move the source overlay files into a `src` directory. Right now the overlay files are being copied to the root of the plugin directory instead of overlaying `src/index.ts` as we would want.

Jup, moving `techdocs` overlay data from `overlay` to `overlay/src` should fix it

gashcrumb (approved):

dzemanov (approved):

PR #2383, opened 2026-05-04
feat: E2E coverage instrumentation with CI pipeline for all dynamic plugins

gustavolira:
## Summary

Adds Istanbul-based E2E coverage collection for dynamic plugins with a CI pipeline that automatically builds instrumented OCI images when `source.json` changes — and skips builds when the image already exists.

**Scope: frontend plugins only.** Coverage collection uses `window.__coverage__` which is only available for browser-executed code. Backend plugin coverage would require a different collection mechanism (`global.__coverage__` from the Node.js process) and is out of scope for this PR.

### How it works

1. **`source.json` changes** (push to main) trigger the `build-instrumented-plugins` workflow
2. For each workspace with `e2e-tests/`, the workflow checks if an instrumented OCI image already exists for the current `repo-ref`
3. If not, it runs `instrument-plugin.sh` which: clones upstream at the exact ref, builds the plugin (`backstage-cli` + `janus-cli export-dynamic`), then applies `nyc instrument` on the final webpack output
4. The instrumented bundle is published as an OCI artifact to `ghcr.io`
5. During E2E tests with `E2E_COLLECT_COVERAGE=1`, the `_coverageCollector` auto-fixture (in [e2e-test-utils PR #95](https://github.com/redhat-developer/rhdh-e2e-test-utils/pull/95)) collects `window.__coverage__` from the browser after each test
6. `report-coverage.sh` merges per-test coverage JSONs via `nyc merge`, generates lcov via `nyc report`
7. `upload-coverage.sh` uploads lcov to Codecov with per-workspace flags (`e2e-tech-radar`, `e2e-topology`, etc.) for dashboard filtering

### Key design decisions

- **Post-webpack instrumentation**: `babel-plugin-istanbul` gets stripped by webpack's module federation during `export-dynamic-plugin`. Applying `nyc instrument` **after** the final webpack build preserves Istanbul `__coverage__` in the browser runtime
- **Caching by source.json ref**: OCI images are tagged with `ref-<sha-first-12>`. If the ref hasn't changed, the build is skipped entirely — no CI overhead
- **Cross-repo Codecov attribution**: Coverage uploads use the upstream repo's SHA and slug from `source.json`, so coverage appears on the correct repo in Codecov
- **Parallel matrix**: All 18 workspaces with E2E tests build in parallel via GitHub Actions matrix strategy
- **`nyc` CLI for merge/report**: Coverage merging and lcov generation use `nyc merge` + `nyc report` instead of custom code — `nyc` is already a pipeline dependency and handles lcov edge cases better
- **`testInfo.project.outputDir` for coverage path**: The auto-fixture writes coverage JSONs to `<outputDir>/coverage/`, avoiding CWD mismatch issues between Playwright workers and the main process

### Known limitations

- **Overlays and patches are not applied**: The instrumented build clones upstream source directly. Production builds apply overlays from `workspaces/*/plugins/*/overlay/` and patches from `workspaces/*/patches/*.patch`. Most workspaces have no overlays/patches, so the instrumented build matches production in the majority of cases. Tracking as a follow-up.
- **Single frontend plugin per workspace**: The build discovers the first `frontend-plugin` in metadata. Workspaces with multiple frontend plugins only instrument the first one.
- **Missing injection integration**: This PR builds instrumented OCI artifacts and provides the collection/upload mechanism, but the logic to swap production plugins with instrumented ones during E2E execution is a follow-up in e2e-test-utils.

### Files

| File | Purpose |
|------|---------|
| `.github/workflows/build-instrumented-plugins.yaml` | CI workflow: detect changed workspaces, check OCI cache, build and publish |
| `scripts/instrument-plugin.sh` | Build instrumented dynamic plugin from upstream source |
| `scripts/report-coverage.sh` | Merge per-test coverage JSONs (nyc merge), generate lcov (nyc report), upload to Codecov |
| `scripts/upload-coverage.sh` | Upload lcov to Codecov with cross-repo attribution and flags |

### Related PRs

- [rhdh-e2e-test-utils PR #95](https://github.com/redhat-developer/rhdh-e2e-test-utils/pull/95) — `_coverageCollector` auto-fixture that collects `window.__coverage__` per test

### Usage

```bash
# Manual: instrument a specific plugin
./scripts/instrument-plugin.sh tech-radar

# Run E2E with coverage collection
E2E_COLLECT_COVERAGE=1 ./run-e2e.sh -w tech-radar

# Merge + upload (called automatically by run-e2e.sh)
./scripts/report-coverage.sh tech-radar

# Upload only (requires lcov already generated)
CODECOV_TOKEN=xxx ./scripts/upload-coverage.sh tech-radar
```

## Test plan

- [ ] Verify workflow detects changed `source.json` files correctly
- [ ] Verify OCI image existence check skips unnecessary builds
- [ ] Run `./scripts/instrument-plugin.sh tech-radar` locally — verify `__coverage__` in output
- [ ] Verify force-rebuild workflow_dispatch input bypasses cache
- [ ] Verify Codecov flags produce per-plugin dashboard filtering

Ref: [RHIDP-13411](https://redhat.atlassian.net/browse/RHIDP-13411)


gustavolira:
/review
-i
--pr_reviewer.require_score_review=true
--pr_reviewer.require_can_be_split_review=true
--pr_reviewer.num_code_suggestions="2"

e2e-coverage/coverage-reporter.ts
```diff
@@ -0,0 +1,155 @@
+/**
+ * Playwright custom reporter that merges Istanbul coverage files
+ * and converts to lcov format for Codecov upload.
+ *
+ * Usage in playwright.config.ts:
+ *   reporter: [['list'], ['../e2e-coverage/coverage-reporter.ts']],
+ *
+ * Requires: E2E_COLLECT_COVERAGE=1
+ */
+
```
subhashkhileri: **Suggestion: Delete this file — `nyc` CLI replaces it entirely.**

This hand-rolls Istanbul merge logic (`mergeCoverage`) and lcov conversion (`coverageToLcov`, ~60 lines) that `nyc` already provides natively — and `nyc` is already part of this pipeline (`nyc instrument`).

Replace with 2 lines in `run-e2e.sh` after tests complete:
```bash
npx nyc merge node_modules/.cache/e2e-test-results/coverage coverage/coverage-final.json
npx nyc report --temp-dir coverage --reporter=lcov --reporter=text-summary --report-dir coverage
```

Benefits:
- ~155 lines of custom code removed
- Battle-tested lcov output (handles edge cases like multi-line functions, implicit else branches)
- No reporter injection in the generated `playwright.config.ts`
- Eliminates a CWD mismatch issue: this reporter runs in the main Playwright process where `process.cwd()` is the repo root, but the fixture writes coverage JSONs from workers where `process.cwd()` is the workspace's `e2e-tests/` dir — they never read from the same directory

e2e-coverage/coverage-utils.ts
```diff
@@ -0,0 +1,59 @@
+import * as path from "node:path";
+
+export const COVERAGE_DIR = path.resolve(
+  process.cwd(),
+  process.env.COVERAGE_OUTPUT_DIR || "coverage/istanbul",
+);
+export const COLLECT_COVERAGE = process.env.E2E_COLLECT_COVERAGE === "1";
```
subhashkhileri: **Suggestion: Delete this file — only consumed by the custom reporter.**

The `CoverageData` types, `mergeCoverage`, and path constants are only used by `coverage-reporter.ts`. If the reporter is replaced by `nyc` CLI (see comment on that file), this file has no consumers.

Also note: `COVERAGE_DIR` and `COLLECT_COVERAGE` are evaluated at module load time, capturing `process.cwd()` at import. In the reporter (main process), CWD is the repo root. In the fixture (worker process), CWD is the workspace dir. This divergence means the reporter and fixture silently target different directories.

run-e2e.sh
```diff
@@ -223,6 +226,12 @@ for ws in "${E2E_WORKSPACES[@]}"; do
     done <<< "$PROJECTS_BLOCK"
 done
 
+COVERAGE_REPORTER_LINE=""
+if [[ "${E2E_COLLECT_COVERAGE:-}" == "1" ]]; then
+    COVERAGE_REPORTER_LINE="  reporter: [...(baseConfig.reporter || []), ['./e2e-coverage/coverage-reporter.ts']],"
```
subhashkhileri: **Suggestion: Remove this reporter injection block.**

If the custom reporter is replaced by `nyc` CLI (post-test), there's no need to inject a reporter into the generated `playwright.config.ts`. The fixture in e2e-test-utils writes coverage JSONs during the test run; `nyc merge` + `nyc report` processes them afterward. The reporter was the bridge between these two — but a shell command after `npx playwright test` does the same thing with less moving parts.

run-e2e.sh
```diff
@@ -265,6 +275,26 @@ echo ""
 TEST_EXIT_CODE=0
 npx playwright test "${PLAYWRIGHT_ARGS[@]+"${PLAYWRIGHT_ARGS[@]}"}" || TEST_EXIT_CODE=$?
 
+# ── Upload coverage ──────────────────────────────────────────────────────
+# The merged lcov.info contains coverage from ALL workspaces. Each upload
+# sends the full file with a different --flag and --sha (from source.json).
+# Codecov scopes coverage by --slug, so cross-repo data is ignored.
+if [[ "${E2E_COLLECT_COVERAGE:-}" == "1" ]] && [[ -f "coverage/istanbul/lcov.info" ]]; then
```
subhashkhileri: **Two suggestions here:**

**1. Replace with `nyc` merge + report + upload:**
```bash
COVERAGE_JSON_DIR="node_modules/.cache/e2e-test-results/coverage"

if [[ "${E2E_COLLECT_COVERAGE:-}" == "1" ]] && ls "$COVERAGE_JSON_DIR"/*.json &>/dev/null; then
    npx nyc merge "$COVERAGE_JSON_DIR" coverage/coverage-final.json
    npx nyc report --temp-dir coverage --reporter=lcov --reporter=text-summary --report-dir coverage

    for ws in "${E2E_WORKSPACES[@]}"; do
        if [[ -f "workspaces/$ws/source.json" ]]; then
            "$SCRIPT_DIR/scripts/upload-coverage.sh" "$ws" || \
                echo "[WARN] Coverage upload failed for $ws (non-fatal)"
        fi
    done
fi
```

The coverage JSON path (`node_modules/.cache/e2e-test-results/coverage`) aligns with Playwright's existing `outputDir` convention. No new directory paths introduced.

**2. Gate to `e2e-ocp-helm` PR check job only.**
Coverage collection should only run on the `e2e-ocp-helm` PR check job where instrumented OCI images are available. Nightly jobs use released OCI refs (not instrumented builds), so `window.__coverage__` would be empty — coverage collection would produce no data and the upload would be noise.

scripts/upload-coverage.sh
```diff
@@ -0,0 +1,87 @@
...
+# Optional environment:
+#   COVERAGE_OUTPUT_DIR - Override coverage directory (default: coverage/istanbul)
+
+set -euo pipefail
+
+WORKSPACE="${1:?Usage: $0 <workspace-name>}"
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+REPO_ROOT="$(cd "$SCRIPT_DIR/.." && pwd)"
+WORKSPACE_DIR="$REPO_ROOT/workspaces/$WORKSPACE"
+COVERAGE_DIR="${COVERAGE_OUTPUT_DIR:-$REPO_ROOT/coverage/istanbul}"
```
subhashkhileri: **Suggestion: Remove `COVERAGE_OUTPUT_DIR` — hardcode the path.**

The coverage output location is an internal implementation detail, not a user-facing config. Hardcode it:
```bash
COVERAGE_DIR="$REPO_ROOT/coverage"
LCOV_FILE="$COVERAGE_DIR/lcov.info"
```

This matches where `nyc report --report-dir coverage` writes the lcov file. No env var needed.

subhashkhileri (commented):
### Review: 

the Playwright integration can be significantly simplified by replacing custom code with `nyc` CLI. See inline comments.

**Also:** Coverage collection (`E2E_COLLECT_COVERAGE=1`) should only be enabled on the `e2e-ocp-helm` PR check job — nightly uses released OCI refs (not instrumented builds), so coverage would produce empty data.

run-e2e.sh
gustavolira: @subhashkhileri about the item 2, coverage collection is already gated by `E2E_COLLECT_COVERAGE=1`. The CI job definition controls when to set that env var, so the gating happens at the job level, not in run-e2e.sh. Nightly jobs simply won't set the flag.

scripts/instrument-plugin.sh
```diff
@@ -0,0 +1,205 @@
...
+mkdir -p "$CLONE_DIR"
+
+git clone --depth 1 "$REPO_URL" "$CLONE_DIR" 2>&1 || {
+  echo "Shallow clone failed, falling back to full clone" >&2
+  rm -rf "$CLONE_DIR"
+  git clone "$REPO_URL" "$CLONE_DIR"
+}
+cd "$CLONE_DIR"
+git fetch --depth 1 origin "$REPO_REF" 2>/dev/null || git fetch origin "$REPO_REF"
+git checkout "$REPO_REF"
```
subhashkhileri: **Overlays and patches are NOT applied.**

The existing production workflow (`publish-workspace-plugins.yaml` → `export-workspaces-as-dynamic.yaml`) applies overlay files from `workspaces/*/plugins/*/overlay/` and patches from `workspaces/*/patches/*.patch` before building. This script clones upstream and builds directly — no overlays, no patches.

This means the instrumented code **does not match** the production plugin. If an overlay fixes a bug, changes imports, or modifies behavior, the instrumented version won't have those changes. Coverage data would be against a different version of the code than what's actually deployed and tested.

This should either:
1. Apply overlays/patches the same way the production workflow does (preferred — ensures coverage matches production), or
2. Be explicitly documented as a known limitation with a tracking issue

subhashkhileri (commented):
### Review: Instrumentation Pipeline

Following up on the Playwright integration review — this covers the instrumentation workflow and build script.

The post-webpack instrumentation approach (`nyc instrument` after `export-dynamic-plugin`) is the right technique for module federation. However, there are significant gaps between how this workflow builds plugins vs. how the existing production workflow does it.

.github/workflows/build-instrumented-plugins.yaml
```diff
@@ -0,0 +1,271 @@
...
+            echo "ERROR: Instrumented bundle not found at $BUNDLE_DIR"
+            exit 1
+          fi
+
+          TAR_FILE=".instrumented/${WORKSPACE}.tar.gz"
+          tar -czf "$TAR_FILE" -C "$BUNDLE_DIR" .
+
+          oras push "$IMAGE_REF" \
+            --artifact-type "application/vnd.rhdh.dynamic-plugin.coverage.v1" \
+            "$TAR_FILE:application/gzip"
```
subhashkhileri: **Different OCI format from production — how will the test framework consume this?**

The existing production workflow creates OCI **container images** via Podman. This workflow creates OCI **artifacts** (a tar.gz via ORAS with custom artifact type `application/vnd.rhdh.dynamic-plugin.coverage.v1`). These are fundamentally different formats.

The test framework's plugin metadata resolution reads `spec.dynamicArtifact` from metadata YAML, which points to production OCI container images pulled by the RHDH container. How does the test framework know to pull, extract, and serve an ORAS artifact instead? I don't see any integration for consuming instrumented images during E2E test execution.

Is there a missing piece that swaps the production OCI ref with the instrumented one when `E2E_COLLECT_COVERAGE=1`?

.github/workflows/build-instrumented-plugins.yaml
```diff
@@ -0,0 +1,271 @@
...
+          REPO_REF=$(jq -r '.["repo-ref"]' "$SOURCE_JSON")
+          REPO_URL=$(jq -r '.repo' "$SOURCE_JSON")
+          SLUG=$(echo "$REPO_URL" | sed 's|https://github.com/||; s|\.git$||')
+          REF_SHORT="${REPO_REF:0:12}"
+
+          # Find the frontend plugin package name from metadata
+          PLUGIN_IMAGE_NAME=""
+          for meta_file in "workspaces/${WORKSPACE}/metadata"/*.yaml; do
+            [ -e "$meta_file" ] || continue
+            ROLE=$(yq -r '.spec.backstage.role // ""' "$meta_file")
```
subhashkhileri: **Plugin discovery: only finds the first frontend plugin.**

The production workflow uses `plugins-list.yaml` which explicitly lists all plugins to export (there can be multiple per workspace). This code scans metadata files and `break`s on the first `frontend-plugin` match.

For workspaces with multiple frontend plugins (e.g., `backstage` workspace has several), only one gets instrumented. The others silently get no coverage.

Consider iterating over all frontend plugins in the metadata, or at minimum using the same `plugins-list.yaml` that the production workflow uses.

.github/workflows/build-instrumented-plugins.yaml
```diff
@@ -0,0 +1,271 @@
...
+          done
+
+          if [[ -z "$PLUGIN_IMAGE_NAME" ]]; then
+            echo "No frontend plugin found in metadata for ${WORKSPACE}"
+            echo "skip=true" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
+
+          IMAGE_REPO="ghcr.io/${GITHUB_REPO}/${PLUGIN_IMAGE_NAME}-coverage"
+          IMAGE_TAG="ref-${REF_SHORT}"
```
subhashkhileri: **Tag format: `ref-{first-12-chars}` can collide when `repo-ref` is a tag name.**

`REF_SHORT="${REPO_REF:0:12}"` on line 131 takes the first 12 characters of `repo-ref`. If `repo-ref` is a git tag like `v1.2.3`, `REF_SHORT` is just `v1.2.3` (7 chars, no truncation). This works but is fragile:

- Short tag names from different workspaces sharing the same upstream repo could collide (e.g., two workspaces both pointing to the same tag)
- The tag format diverges from production tags which encode semantic versioning (`bs_1.49.4__2.8.0`)

Consider hashing the full ref: `REF_SHORT=$(echo -n "$REPO_REF" | sha256sum | head -c 12)` — this guarantees uniqueness regardless of ref format.

scripts/instrument-plugin.sh
```diff
@@ -0,0 +1,205 @@
...
+echo "--- Step 3: Finding frontend plugin package ---"
+if [[ -n "$PLUGIN_NAME" ]]; then
+  PLUGIN_PKG_DIR=$(find "$PLUGIN_WORKSPACE_DIR" -name "package.json" -path "*/$PLUGIN_NAME/*" -not -path "*/node_modules/*" | head -1 | xargs dirname)
+else
+  PLUGIN_PKG_DIR=$(find "$PLUGIN_WORKSPACE_DIR" -name "package.json" -not -path "*/node_modules/*" -not -path "*/e2e-*/*" -not -path "*/backend*/*" -not -path "*/module-*/*" -not -path "$PLUGIN_WORKSPACE_DIR/package.json" | while read -r pkg; do
+    if jq -e '.backstage.role == "frontend-plugin"' "$pkg" >/dev/null 2>&1; then
+      dirname "$pkg"
+      break
+    fi
+  done)
```
subhashkhileri: **Only frontend plugins are instrumented — backend coverage is out of scope.**

The script filters for `backstage.role == "frontend-plugin"` (and the workflow does the same in metadata scanning). This is correct for browser-based `window.__coverage__` collection, but worth noting explicitly:

- Backend plugins (`backend-plugin`, `backend-plugin-module`) are excluded
- If backend coverage is ever needed, it would require a different collection mechanism (e.g., Node.js `global.__coverage__` extraction from the RHDH container)

Not blocking — just flag it in the PR description so it's clear this is frontend-only by design, not by omission.

subhashkhileri (commented):
## Proposal: Reuse the existing build pipeline instead of a separate instrumentation workflow

The current PR introduces `scripts/instrument-plugin.sh` (205 lines) and `.github/workflows/build-instrumented-plugins.yaml` (271 lines) — a parallel pipeline that re-clones, re-builds, and re-packages plugins from scratch. This duplicates the production build pipeline and misses overlays, patches, and edge cases that the production pipeline already handles.

### Proposed approach

Instead of rebuilding from source, **instrument the already-published production image**. Add an `instrument` job to `publish-release-branch-workspace-plugins.yaml` that runs after the `export` job:

```yaml
  instrument:
    needs: export
    if: # only for workspaces that have e2e-tests/
    runs-on: ubuntu-latest
    strategy:
      matrix:
        workspace: # detect workspaces with e2e-tests/ (similar to current detect-workspaces job)
      fail-fast: false
    permissions:
      contents: read
      packages: write
    steps:
      - uses: actions/checkout@v6

      - name: Resolve published image ref
        id: meta
        run: |
          # Read source.json + metadata to find the frontend plugin's image name and tag
          # Use the SAME naming/tagging as production: ghcr.io/{repo}/{plugin}:{tag}
          # The production image was just pushed by the export job

      - name: Log in to GHCR
        uses: docker/login-action@v4
        with:
          registry: ghcr.io
          username: ${{ github.actor }}
          password: ${{ secrets.GITHUB_TOKEN }}

      - name: Setup Node.js
        uses: actions/setup-node@v6

      - name: Instrument and publish coverage image
        run: |
          PROD_IMAGE="${{ steps.meta.outputs.image-ref }}"
          COVERAGE_IMAGE="${{ steps.meta.outputs.coverage-image-ref }}"

          # 1. Extract plugin bundle from production image
          CONTAINER=$(podman create "$PROD_IMAGE")
          mkdir -p .instrumented
          podman cp "$CONTAINER:/opt/app-root/src/" .instrumented/plugin/
          podman rm "$CONTAINER"

          # 2. Instrument JS files with Istanbul
          npx --yes nyc instrument .instrumented/plugin/ .instrumented/instrumented/ --source-map

          # 3. Build new image from instrumented files
          cat > .instrumented/Containerfile <<EOF
          FROM scratch
          COPY instrumented/ /opt/app-root/src/
          EOF
          podman build -t "$COVERAGE_IMAGE" -f .instrumented/Containerfile .instrumented/

          # 4. Push with -coverage suffix, same tag as production
          podman push "$COVERAGE_IMAGE"
```

### What this eliminates

| File | Lines | Reason |
|------|-------|--------|
| `scripts/instrument-plugin.sh` | 205 | No need to re-clone, re-build, re-export from source |
| `.github/workflows/build-instrumented-plugins.yaml` | 271 | No separate workflow needed |

### What this fixes

1. **Overlays and patches are included** — the production image already has them applied
2. **All plugins discovered** — not just the first `frontend-plugin` match
3. **Same toolset** — Podman throughout, matching the production pipeline (no ORAS)
4. **Same tag format** — `bs_1.49.4__1.32.0` with `-coverage` suffix on the image name
5. **Same triggers** — runs whenever production images are published (push to release branches)
6. **No pnpm/npm detection needed** — the plugin is already built

### Coverage image naming

Production: `ghcr.io/{repo}/{plugin}:{tag}`
Coverage: `ghcr.io/{repo}/{plugin}-coverage:{tag}`

Same tag, `-coverage` suffix on the image name. This makes it trivial to swap in the coverage image during E2E runs — just append `-coverage` to the image name.

### Note on image structure

The `podman cp` + `podman build FROM scratch` approach needs to match whatever directory layout the `rhdh-cli package` command produces. The exact path inside the container (e.g., `/opt/app-root/src/`) needs verification against the actual production image. The Containerfile above is illustrative — the actual paths should be confirmed by inspecting a published production image with `podman inspect` or `skopeo inspect`.

An alternative simpler approach: instead of rebuilding from a Containerfile, use `podman commit` after modifying the extracted files in-place, or use `buildah` to add/replace layers. The goal is to match the production image format exactly.

subhashkhileri (commented):
## Move coverage logic out of `run-e2e.sh`

The ~30 lines of coverage code added to `run-e2e.sh` (nyc merge, nyc report, multi-workspace warning, per-workspace upload loop) should move into a self-contained script — e.g. rename `scripts/upload-coverage.sh` → `scripts/report-coverage.sh` and have it handle the full pipeline (merge → report → upload).

`run-e2e.sh` then reduces to:
```bash
if [[ "${E2E_COLLECT_COVERAGE:-}" == "1" ]]; then
    "$SCRIPT_DIR/scripts/report-coverage.sh" "${E2E_WORKSPACES[@]}"
fi
```

Keeps `run-e2e.sh` focused on test orchestration, makes the coverage script independently re-runnable (useful for debugging uploads without re-running tests), and avoids splitting the logic across two files.

run-e2e.sh
```diff
@@ -265,6 +272,33 @@ echo ""
 TEST_EXIT_CODE=0
 npx playwright test "${PLAYWRIGHT_ARGS[@]+"${PLAYWRIGHT_ARGS[@]}"}" || TEST_EXIT_CODE=$?
 
+# ── Merge and upload coverage ────────────────────────────────────────────
+if [[ "${E2E_COLLECT_COVERAGE:-}" == "1" ]]; then
+    COVERAGE_JSON_DIR="node_modules/.cache/e2e-test-results/coverage"
+    if ls "$COVERAGE_JSON_DIR"/*.json &>/dev/null; then
```
subhashkhileri: This entire block (merge → report → warning → upload loop) can move into a self-contained script. Rename `scripts/upload-coverage.sh` → `scripts/report-coverage.sh` and have it handle the full pipeline. Then this becomes:

```bash
if [[ "${E2E_COLLECT_COVERAGE:-}" == "1" ]]; then
    "$SCRIPT_DIR/scripts/report-coverage.sh" "${E2E_WORKSPACES[@]}"
fi
```

Keeps `run-e2e.sh` focused on test orchestration, makes the coverage script independently re-runnable (useful for debugging uploads without re-running tests), and avoids splitting logic across two files.

scripts/instrument-plugin.sh
gustavolira: @subhashkhileri Good catch. Most workspaces have no overlays or patches, so the instrumented build matches production for the majority of cases. For the few that do, the coverage delta would be small since overlays typically add/replace config files rather than modify plugin logic. Tracking this as a known limitation for v1 — we can add overlay/patch application in a follow-up once the basic pipeline is validated end-to-end.

.github/workflows/build-instrumented-plugins.yaml
gustavolira: Correct — the integration piece that swaps the production plugin with the instrumented one during E2E execution is not in this PR. This PR establishes the build pipeline (instrumented OCI artifacts) and collection mechanism (auto-fixture + nyc merge + Codecov upload). The injection logic — pulling the ORAS artifact and mounting it into the RHDH container when E2E_COLLECT_COVERAGE=1 — is planned as a follow-up in e2e-test-utils

.github/workflows/build-instrumented-plugins.yaml
gustavolira: Intentional for v1 — each workspace's E2E tests typically exercise one primary frontend plugin. Multi-plugin instrumentation would require iterating over metadata, building each plugin separately, and publishing multiple OCI images per workspace. Can be added as a follow-up if there's a concrete workspace that needs it.

.github/workflows/build-instrumented-plugins.yaml
gustavolira: Each workspace produces a differently-named image (<plugin-name>-coverage), so two workspaces pointing to the same tag don't collide — they're different packages. For SHAs, 12 chars gives ~2^48 collision resistance. Hashing would work but loses readability — ref-v1.2.3 is immediately useful in debugging, ref-a1b2c3d4e5f6 is not. Keeping as-is.

scripts/instrument-plugin.sh
gustavolira: Correct, by design. window.__coverage__ is only available for browser-executed code. Backend coverage would require extracting global.__coverage__ from the RHDH Node.js process, which is a fundamentally different collection mechanism. Will add a note to the PR description clarifying this is frontend-only scope.

.github/workflows/build-instrumented-plugins.yaml
```diff
@@ -0,0 +1,271 @@
...
+
+          if [[ ! -d "$BUNDLE_DIR" ]]; then
+            echo "ERROR: Instrumented bundle not found at $BUNDLE_DIR"
+            exit 1
+          fi
+
+          TAR_FILE=".instrumented/${WORKSPACE}.tar.gz"
+          tar -czf "$TAR_FILE" -C "$BUNDLE_DIR" .
+
+          oras push "$IMAGE_REF" \
```
kadel: This uses completely different way of packaging plugins than what production artifact uses.

The whole process of rebuilding plugins for testing seems looks problematic to me. I'm afraid that it will create a lot of problems down the line. It will be really easy to get those two ways of building out of sync (they already are) which will result in testing something different from what we ship. 


We should consider alternative. I've tried to describe it in https://github.com/redhat-developer/rhdh-plugin-export-overlays/pull/2383#discussion_r3233125937

scripts/instrument-plugin.sh
kadel, line 1: This re-implements the full build pipeline that already exists in the production OCI image published by rhdh-plugin-export-utils. The only extra is `nyc instrument`.

What instead of reimplementing the build process and rebuilding we re-use existing artifacts.
We could pull the production image, extract it, add instrumentation and package it back to image with coverage. This way we don't rebuild the whole plugin. And minimize the difference between what is tested and what is shipped. 

Something like this:
```
OCI_REF=ghcr.io/redhat-developer/rhdh-plugin-export-overlays/backstage-community-plugin-quay:next__1.28.1
COVERAGE_IMAGE_REF=ghcr.io/redhat-developer/rhdh-plugin-export-overlays/backstage-community-plugin-quay-coverage:next__1.28.1

# Pull production image
podman pull "$OCI_REF"

# Create a container (don't start it)
CID=$(podman create "$OCI_REF")

# Copy static/ out, instrument, copy back
podman cp "$CID:backstage-community-plugin-quay/dist/static" ./static-original/
npx nyc instrument ./static-original ./static-instrumented --source-map
podman cp ./static-instrumented/static "$CID:/backstage-community-plugin-quay/dist/"

# Commit as new image and push
podman commit "$CID" "$COVERAGE_IMAGE_REF"
podman rm "$CID"
# podman push "$COVERAGE_IMAGE_REF"
```

kadel (changes requested):

scripts/instrument-plugin.sh
gustavolira: @kadel  Much better approach! Updated

psrna:
@gustavolira @kadel Are there any other areas where you don't have consensus on the approach? Has all the requested changes/feedback been addressed? What is still outstanding? Thanks!

scripts/upload-coverage.sh
```diff
@@ -0,0 +1,131 @@
...
+    darwin) CODECOV_OS="macos" ;;
+    *)
+      echo "ERROR: Unsupported OS: $OS" >&2
+      exit 1
+      ;;
+  esac
+
+  echo ""
+  echo "Downloading Codecov CLI for ${CODECOV_OS}..."
+  curl -sL -o "$CODECOV_BIN" "https://cli.codecov.io/latest/${CODECOV_OS}/codecov"
```
kadel: Wouldn't it be safer to pin to a specific version, rather than always using latest?

kadel:
> @gustavolira @kadel Are there any other areas where you don't have consensus on the approach? Has all the requested changes/feedback been addressed? What is still outstanding? Thanks!

I think that we have consensus on overall approach. But the code is not ready to be merged. There are still some issues with code.

@psrna I'm happy to help with the review. But I think that someone who knows more about test setup in this repo (like @subhashkhileri or @zdrapela) should have the final "ack"

scripts/report-coverage.sh
```diff
@@ -0,0 +1,56 @@
...
+if [[ $# -eq 0 ]]; then
+  echo "Usage: $0 <workspace> [workspace...]" >&2
+  exit 1
+fi
+
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+REPO_ROOT="$(cd "$SCRIPT_DIR/.." && pwd)"
+WORKSPACES=("$@")
+
+COVERAGE_JSON_DIR="${COVERAGE_OUTPUT_DIR:-node_modules/.cache/e2e-test-results/coverage}"
```
kadel: This was previously raised by @subhashkhileri 
`COVERAGE_OUTPUT_DIR` is never defined or exported in this script

scripts/report-coverage.sh
```diff
@@ -0,0 +1,56 @@
...
+
+if [[ ${#WORKSPACES[@]} -gt 1 ]]; then
+  echo "[WARN] Coverage data is merged across all ${#WORKSPACES[@]} workspaces."
+  echo "[WARN] For clean per-workspace coverage, run with a single -w flag."
+fi
+
+echo "[INFO] Uploading E2E coverage to Codecov..."
+for ws in "${WORKSPACES[@]}"; do
+  if [[ -f "$REPO_ROOT/workspaces/$ws/source.json" ]]; then
+    "$SCRIPT_DIR/upload-coverage.sh" "$ws" || \
```
kadel: If this was run with multiple workspaces, and they were previously merged into one coverage data, won't this mean the merged coverage data gets uploaded into each workspace?
Does Codecov handle this correctly? Does it ignore file paths that don't belong to the source repo? Have you tested a multi-workspace upload to verify?

scripts/upload-coverage.sh
```diff
@@ -0,0 +1,131 @@
...
+  --file "$LCOV_FILE" \
+  --flag "e2e-$WORKSPACE" \
+  --sha "$REPO_REF" \
+  --slug "$SLUG" \
+  --token "$CODECOV_TOKEN" \
+  --git-service github \
+  --name "overlay-e2e-$WORKSPACE" \
+  --disable-search \
+  --fail-on-error || {
+    echo "[WARN] Codecov upload failed (non-fatal)"
```
kadel: why is failed upload non-fatal? Isn't the whole point of this workflow to submit coverage report to codecov?

kadel (changes requested):

scripts/upload-coverage.sh
gustavolira: Pinned codecov-cli to version 9.0.4

scripts/report-coverage.sh
gustavolira: @kadel this file has been removed.T he coverage merge and lcov generation is now handled by the Playwright CoverageReporter (in e2e-coverage/coverage-reporter.ts), which runs as part of the test process itself, no separate shell script needed. The COVERAGE_OUTPUT_DIR issue no longer applies.

scripts/report-coverage.sh
gustavolira: This was indeed broken for multi-workspace runs. 
Upload is handled separately by `upload-coverage.sh`, which takes a single workspace argument and uploads the lcov for that workspace only. For multi-workspace runs, the caller is responsible for invoking upload once per workspace with workspace-scoped coverage data.

scripts/upload-coverage.sh
gustavolira: You're right, now the script runs under set -euo pipefail, so any failure in codecov upload-process now terminates the script with a non-zero exit code

.github/workflows/build-instrumented-plugins.yaml
subhashkhileri, line 1: **1). I think we might not need build-instrumented-plugins.yaml** as a separate workflow. It triggers on source.json changes to main, but the instrumented images are actually consumed during the PR flow , after /publish and before E2E tests. The existing auto-publish-pr.yaml already has everything we need: published-exports gives us the exact image refs just pushed, and workspace metadata tells us which are frontend plugins.

What if we just add an instrument job in that existing flow?

```
/publish → export → instrument (new) → add-completion-comment → "/test e2e-ocp-helm"
                        │
                        └─ reads published-exports, filters for frontend plugins if required,
                           instruments and, pushes with __coverage tag
```

No separate triggers, no GHCR existence checks, no workspace detectio, the publish flow already handles all of that.

**2). On naming, can we use tag suffix instead of image name suffix?** The PR creates separate GHCR packages (plugin-coverage:tag). Since we already use composite tags like bs_1.49.4__1.5.0, appending __coverage  makes the image swap a simple.

**3). Nothing actually consumes these instrumented images yet.** We should update rhdh-e2e-test-utils to handle the PR image swap for coverage images when E2E_COLLECT_COVERAGE=1. Might be worth getting that wired up before merging so we can validate the full pipeline end-to-end.

gustavolira: > **1). I think we might not need build-instrumented-plugins.yaml** as a separate workflow. It triggers on source.json changes to main, but the instrumented images are actually consumed during the PR flow , after /publish and before E2E tests. The existing auto-publish-pr.yaml already has everything we need: published-exports gives us the exact image refs just pushed, and workspace metadata tells us which are frontend plugins.
> 
> What if we just add an instrument job in that existing flow?
> 
> ```
> /publish → export → instrument (new) → add-completion-comment → "/test e2e-ocp-helm"
>                         │
>                         └─ reads published-exports, filters for frontend plugins if required,
>                            instruments and, pushes with __coverage tag
> ```
> 
> No separate triggers, no GHCR existence checks, no workspace detectio, the publish flow already handles all of that.
> 
> **2). On naming, can we use tag suffix instead of image name suffix?** The PR creates separate GHCR packages (plugin-coverage:tag). Since we already use composite tags like bs_1.49.4__1.5.0, appending __coverage makes the image swap a simple.
> 
> **3). Nothing actually consumes these instrumented images yet.** We should update rhdh-e2e-test-utils to handle the PR image swap for coverage images when E2E_COLLECT_COVERAGE=1. Might be worth getting that wired up before merging so we can validate the full pipeline end-to-end.

Thanks for the suggestions, @subhashkhileri .

1. Merging into auto-publish-pr.yaml: These workflows serve different purposes. build-instrumented-plugins.yaml pre-builds instrumented images from production OCI artifacts on main — it's a cache-warming step so coverage images are ready before any E2E run. auto-publish-pr.yaml builds PR-specific images from the overlay branch. Merging them would mean coverage images are only available after /publish on a PR, losing the pre-caching benefit. The whole point is that instrumentation is applied to the exact artifact shipped to users, not rebuilt from source during PR flow.
2. Tag suffix vs. image name suffix: Keeping coverage images as separate GHCR packages (plugin-name-coverage:tag) makes registry management cleaner — easier to list, delete, or set permissions independently. A tag suffix (tag__coverage) would mix them into the same package. Either approach works, but separate packages give us better isolation
3. Wiring up consumption before merging: Agreed this is needed, but it's already scoped as a follow-up — the PR description lists "Missing injection integration" as a known limitation and points to e2e-test-utils PR [#95](https://github.com/redhat-developer/rhdh-e2e-test-utils/pull/95). The instrumented images need to exist before the consumer logic can be validated end-to-end. Merging the build pipeline first avoids a circular dependency.

CC: @psrna @kadel

kadel:
RE 1.:

We should have instrumented images for both (build from main/release branch, and PR builds as well).

What @subhashkhileri is suggesting makes sense for PR builds, which we should also do. But it should be separate PR. 

**important question:**
If this PR adds instrumentation to the "release" builds, when it will be used? I don't see any nightly tests or anything that would run e2e against release images. https://github.com/redhat-developer/rhdh-e2e-test-utils/pull/95 doesn't solve this

There is also **RACE CONDITION** in the current approach:

Both `publish-workspace-plugins.yaml` and `build-instrumented-plugins.yaml` trigger on the same push to main.
The `publish` workflow will most likely take longer than ``build-instrumented` workflow.
This will result in `build-instrumented` running on the previous version, not on the lastes one.

gustavolira:
> RE 1.:
> 
> We should have instrumented images for both (build from main/release branch, and PR builds as well).
> 
> What @subhashkhileri is suggesting makes sense for PR builds, which we should also do. But it should be separate PR.
> 
> **important question:** If this PR adds instrumentation to the "release" builds, when it will be used? I don't see any nightly tests or anything that would run e2e against release images. [redhat-developer/rhdh-e2e-test-utils#95](https://github.com/redhat-developer/rhdh-e2e-test-utils/pull/95) doesn't solve this
> 
> There is also **RACE CONDITION** in the current approach:
> 
> Both `publish-workspace-plugins.yaml` and `build-instrumented-plugins.yaml` trigger on the same push to main. The `publish` workflow will most likely take longer than ``build-instrumented`workflow. This will result in`build-instrumented` running on the previous version, not on the lastes one.

@kadel **Race condition is fixed**.
`build-instrumented-plugins.yaml` now uses `workflow_run` instead of a push trigger, so it only runs after “Publish RHDH Release Dynamic Plugin Images” completes successfully on main.

The instrumentation job now reads `workflow_run.head_sha` to check which `source.json` files changed in the publish commit. That way, it always instruments the exact version that was just published, avoiding the race condition we had before.

**Consumer side is added.**
The new `e2e-coverage-nightly.yaml` workflow runs on weekdays or manually. It finds workspaces with a published -coverage image, runs E2E with coverage enabled, and uploads lcov to Codecov.

The only missing piece is the image swap during deployment, which depends on e2e-test-utils PR #95. Until that lands, the workflow still runs E2E but won’t generate coverage data; `run-e2e.sh` now handles this gracefully.

Also, instrument-plugin.sh does not rebuild from source. It instruments the final JS bundles from the already-published production OCI image, so it stays independent of the build process.

subhashkhileri:
> RE 1.:
> 
> We should have instrumented images for both (build from main/release branch, and PR builds as well).
> 
> What @subhashkhileri is suggesting makes sense for PR builds, which we should also do. But it should be separate PR.

@kadel
I have missed from initial reviews that its doing the on main pushes instrumentation(After PR merge).

My understanding was that the scope here is PR checks only, that's where coverage is most actionable (during code review, against the code being changed).
@psrna 

For nightly, I'm not sure it adds much value, it would produce the same coverage data since it's the same plugin code and same tests that already ran during the PR check.

Also, if we did want nightly coverage, then it should instrument the downstream plugin build. Nightly uses registry.access.redhat.com for RHDH-supported plugins and ghcr.io only for community plugins. So we'd need to pull from RHEC/quay, instrument, and push to original source.

On top of that, nightly uses {{inherit}} for supported plugins, which resolves OCI tags through the catalog index images DPDY, Instrumented images with a __coverage tag wouldn't be recognized by {{inherit}} and we'd either need a separate catalog index image with coverage refs, or update the install-dynamic-plugins script to handle coverage variants.

The current PR only triggers on main pushes (After PR merge) and instruments ghcr.io bs_ tagged images, but those aren't what nightly tests actually consume for [supported plugins](https://github.com/redhat-developer/rhdh-plugin-export-overlays/blob/main/default.packages.yaml).

psrna:
@gustavolira @subhashkhileri @kadel Just want to confirm what Subhash said above. The main scope for the E2E coverage was always computing coverage for PR Checks. This was explicitly stated in the requirements in the jira. Thank you!

kadel:
> My understanding was that the scope here is PR checks only, that's where coverage is most actionable (during code review, against the code being changed).

TBH, I did not check what was the original scope. I was only reviewing technical side of the implementation. 

I agree that the coverage on PRs is more useful, and if this is goal than this PR completely misses that.

psrna:
@gustavolira please address the feedback and update the pull request to compute the coverage data on PR checks.
Based on my recent side conversations there seems to be agreement on the approach for instrumenting the artifacts via the current `instrument-plugin.sh` script. Please keep that approach.

Additionally, if you are heavily relying on AI generated code, please remember you remain responsible for that part. I have seen comments indicating a few fixes, but the latest commits in this pull request do not reflect them. The last commit is from one week ago. Also please make sure you personally engage with the review feedback rather than relying on automated tools to produce responses. Even if you use AI assistance, the understanding, verification, and implementation of the changes need to be done and owned by you. Thank you!

.github/workflows/auto-publish-pr.yaml
```diff
@@ -23,11 +23,20 @@ jobs:
       statuses: write
 
     steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+
+      - name: Detect modified workspaces
+        id: detect
+        uses: ./.github/actions/detect-modified-workspaces
```
kadel: this action doesn't exist

scripts/upload-coverage.sh
kadel: this still doesn't answer the question or solve the problem. 

If upload failed than this does `exit 0` so no-error. Is this intentional, are we OK with upload-coverage silently failing?

kadel (changes requested):

.github/workflows/auto-publish-pr.yaml
```diff
@@ -304,6 +306,149 @@ jobs:
...
+            # Check if this is a frontend plugin (only frontend plugins need instrumentation)
+            PLUGIN_ROLE=$(yq -r '.spec.backstage.role // ""' "$METADATA_FILE")
+            if [[ "$PLUGIN_ROLE" != "frontend-plugin" ]]; then
+              echo "  Skipping $PLUGIN_ROLE (only frontend plugins need browser coverage)"
+              continue
+            fi
+
+            # Extract plugin path from dynamicArtifact (after the ! separator)
+            DYNAMIC_ARTIFACT=$(yq -r '.spec.dynamicArtifact // ""' "$METADATA_FILE")
+            if [[ "$DYNAMIC_ARTIFACT" =~ !(.+)$ ]]; then
```
kadel: the `!` is optional, some plugins don't use it even if there is oci artifact  like https://github.com/redhat-developer/rhdh-plugin-export-overlays/blob/main/workspaces/orchestrator/metadata/rhdh-bsp-orchestrator.yaml

kadel: each oci artifact has `io.backstage.dynamic-packages` that has base64 encoded info about the plugins inside the image, you can use that to get the plugin name (directory)

scripts/upload-coverage.sh
```diff
@@ -0,0 +1,155 @@
...
+if [[ -z "${CODECOV_TOKEN:-}" ]]; then
+  echo ""
+  echo "[WARN] CODECOV_TOKEN is not set — skipping Codecov upload"
+  echo "[INFO] Coverage report is still available locally at: $LCOV_FILE"
+  exit 0
+fi
+
+# Download Codecov CLI binary with SHA256 verification.
+# Uses the standalone Go binary (not pip codecov-cli) for supply-chain safety.
+CODECOV_VERSION="v0.7.5"
```
kadel: the latest cli version is v11.2.8, is there a reason to use this old unsupported version?

.github/workflows/auto-publish-pr.yaml
```diff
@@ -304,6 +306,149 @@ jobs:
...
+        with:
+          node-version: 20
+
+      - name: Instrument and publish coverage images
+        if: steps.check-e2e.outputs.has-e2e == 'true'
+        env:
+          PUBLISHED_EXPORTS: ${{ needs.export.outputs.published-exports }}
+          WORKSPACE: ${{ needs.prepare.outputs.workspace }}
+        run: |
+          set -euo pipefail
```
kadel: this is going to be hell to maintain, it might be better to extract this bash script to a file and just call it here. 
This will also make it possible to test separately

scripts/upload-coverage.sh
gustavolira: This is intentional, coverage upload is informational, CI shouldn't fail if the upload wasn't done
I just added a comment explaining [https://github.com/redhat-developer/rhdh-plugin-export-overlays/pull/2383/changes/c2e22bfcf3820a014d91880d47bb0d3cfb2ee95d#diff-1faf8b[…]8d85d0fR160](https://github.com/redhat-developer/rhdh-plugin-export-overlays/pull/2383/changes/c2e22bfcf3820a014d91880d47bb0d3cfb2ee95d#diff-1faf8b483cc585404e800c27cd1d4595e0a2611bf18fb4f14cac61a4a8d85d0fR160)

.github/workflows/auto-publish-pr.yaml
```diff
@@ -304,6 +306,152 @@ jobs:
...
+            [[ -z "$PROD_IMAGE" ]] && continue
+
+            echo "--- Processing: $PROD_IMAGE ---"
+
+            # Extract plugin name from image ref
+            PLUGIN_NAME=$(basename "${PROD_IMAGE%%:*}")
+            echo "  Plugin: $PLUGIN_NAME"
+
+            # Find metadata file for this plugin
+            METADATA_FILE=$(find "${WORKSPACE}/metadata" -name "*.yaml" -exec grep -l "packageName: ${PLUGIN_NAME}" {} \; | head -1 || true)
```
kadel: `packageName` has the format of npm package, for example `@backstage-community/plugin-adr` but `${PLUGIN_NAME}` values is taken from the image so it will have format like `backstage-community-plugin-adr`, so this will never match

scripts/upload-coverage.sh
gustavolira: updated

.github/workflows/auto-publish-pr.yaml
gustavolira: extracted to scripts/instrument-plugin.sh

.github/workflows/auto-publish-pr.yaml
gustavolira: fixed! tyvm!

scripts/instrument-plugin.sh
```diff
@@ -0,0 +1,195 @@
...
+    ((SKIPPED_COUNT++))
+    continue
+  fi
+
+  podman rm "$CID"
+
+  # Instrument with nyc (pinned version for reproducibility)
+  # Must run from work directory to avoid "outside project root" errors
+  echo "  Instrumenting with Istanbul/nyc..."
+  if ! (cd "$WORK_DIR" && npx --yes nyc@15.1.0 instrument dist-original dist-instrumented --source-map); then
```
kadel: latest version is[18.0.0](https://www.npmjs.com/package/nyc/v/18.0.0) 15.1.0 is 6 years old

scripts/instrument-plugin.sh
```diff
@@ -0,0 +1,195 @@
...
+  # Generate coverage image tag: append __coverage suffix to tag
+  # Example: plugin:pr_123__1.2.3 → plugin:pr_123__1.2.3__coverage
+  IMAGE_BASE="${PROD_IMAGE%:*}"
+  IMAGE_TAG="${PROD_IMAGE##*:}"
+  COVERAGE_IMAGE="${IMAGE_BASE}:${IMAGE_TAG}__coverage"
+
+  if ! podman build -t "$COVERAGE_IMAGE" -f "$WORK_DIR/Containerfile" "$WORK_DIR"; then
+    echo "  ❌ Failed to build coverage image - skipping"
+    rm -rf "$WORK_DIR"
+    ((SKIPPED_COUNT++))
```
kadel: This is not going to work as you expect, when `SKIPPED_COUNT` is 0 this will make this script exit with error. 
It is safer to do 

```
SKIPPED_COUNT=$((SKIPPED_COUNT + 1))
```

kadel (approved):

PR #2387, opened 2026-05-05
chore(orchestrator): add translationResources to default Package metadata

lokanandaprabhu:
Updates `workspaces/orchestrator/metadata/rhdh-bsp-orchestrator.yaml `so the default appConfigExamples dynamic plugin entry includes translationResources (Alpha module / orchestratorTranslations), matching the orchestrator plugin’s upstream app-config.yaml after [redhat-developer/rhdh-plugins#2987](https://github.com/redhat-developer/rhdh-plugins/pull/2987).


karthikjeeyar (approved):
/approve
/lgtm

PR #2396, opened 2026-05-05
feat(scorecard): add scorecard dependabot to overlays

alizard0:
https://redhat.atlassian.net/browse/RHIDP-13455


dzemanov (approved):

dzemanov:
Dependabot meta added in this PR: https://github.com/redhat-developer/rhdh-plugin-export-overlays/pull/2392

davidfestal:
> Dependabot meta added in this PR: #2392

Shouldn't we then add also add `rhdh-backstage-plugin-scorecard-backend-module-sonarqube` which was also added in the same PR ?

dzemanov:
> Shouldn't we then add also add rhdh-backstage-plugin-scorecard-backend-module-sonarqube which was also added in the same PR ?

Yes, that would be great to add to stay consistent.
The relation is emitted when plugin has `packages` or when package has `partOf` - so the relation is there as the meta that was added for these packages contains `partOf: scorecard`. But yeah, we should add it also for sonarqube.

PR #2397, opened 2026-05-05
chore: add pre-commit hook for e2e code quality checks

zdrapela:
## Summary

Adds a pre-commit hook for E2E code quality checks and extracts the shared check logic into a standalone script so the hook and CI workflow stay in sync.

### Changes

- **`scripts/e2e-code-quality.sh`** — Single source of truth for ESLint, Prettier, and TypeScript checks. Accepts workspace names as arguments. Used by both the hook and CI.
- **`.githooks/pre-commit`** — Detects staged `workspaces/*/e2e-tests/**` files, delegates to the shared script, then chains to any existing user hooks (global `core.hooksPath` or `.git/hooks/`).
- **`.github/workflows/e2e-code-quality.yaml`** — Simplified to workspace discovery + calling the shared script.
- **`.githooks/README.md`** — Setup instructions and hook chaining docs.
- **`CLAUDE.md`** — Added Developer Setup section with hook setup instructions.

### Setup

```bash
git config core.hooksPath .githooks
```


scripts/e2e-code-quality.sh
```diff
@@ -0,0 +1,121 @@
...
+    echo "ESLint: passed"
+  else
+    report_error "ESLint failed for ${WORKSPACE}"
+    LINT_RESULT="FAIL"
+    FAILED=1
+  fi
+
+  echo ""
+  echo "--- Prettier ---"
+  if (cd "$E2E_DIR" && yarn prettier:check 2>&1); then
```
subhashkhileri: For pre-commit hook we should do auto prettier:fix instead of check. check is good for CI.

scripts/e2e-code-quality.sh
```diff
@@ -0,0 +1,121 @@
...
+    continue
+  fi
+
+  LINT_RESULT="pass"
+  PRETTIER_RESULT="pass"
+  TSC_RESULT="pass"
+
+  echo ""
+  echo "--- ESLint ---"
+  if (cd "$E2E_DIR" && yarn lint:check 2>&1); then
```
subhashkhileri: for pre-commit hook lets replace lint:check with lint:fix.

subhashkhileri (changes requested):

scripts/e2e-code-quality.sh
zdrapela: Nice catch. I added that. It also performs `git add` on the fixed files now.

subhashkhileri (approved):
/lgtm

.githooks/pre-commit
```diff
@@ -0,0 +1,85 @@
...
+  if [[ ${#WORKSPACES[@]} -gt 0 ]]; then
+    echo "E2E Code Quality: fixing and checking staged changes"
+    echo ""
+
+    # Stash unstaged changes so fixers only see staged content.
+    # This protects partially staged files from being fully staged.
+    STASH_NEEDED=false
+    if ! git diff --quiet; then
+      git stash push --keep-index --quiet -m "pre-commit hook backup"
+      STASH_NEEDED=true
```
kadel, line 55: The stash push/pop lifecycle is not protected against interruption. If the user presses Ctrl+C during the fixer run, or if `xargs git add` (line 52) fails under `set -e`, the script aborts before `stash pop` — orphaning the user's unstaged changes in the stash with no recovery guidance.

Consider adding a `trap` to guarantee stash restoration:
```bash
STASH_NEEDED=false
cleanup() {
  if [[ "$STASH_NEEDED" == true ]]; then
    echo "Restoring unstaged changes from stash..." >&2
    git stash pop --quiet 2>/dev/null || {
      echo "WARNING: Could not auto-restore stash. Run 'git stash pop' manually." >&2
    }
    STASH_NEEDED=false
  fi
}
trap cleanup EXIT
```
This single change would also cover the `stash pop` merge conflict case (line 56) and the `xargs git add` failure case (line 52).

.githooks/pre-commit
```diff
@@ -0,0 +1,85 @@
...
+    fi
+
+    # Run fixers; capture exit code to ensure stash is restored on failure
+    FIX_EXIT=0
+    "${REPO_ROOT}/scripts/e2e-code-quality.sh" --fix "${WORKSPACES[@]}" || FIX_EXIT=$?
+
+    # Re-stage files that were auto-fixed
+    echo "$CHANGED_FILES" | xargs git add
```
kadel: This re-stages files unconditionally — even when the fix script failed (e.g. TypeScript errors after ESLint/Prettier auto-fixed files). The user ends up with unreviewed partial fixes staged in their index.

Consider only re-staging on success:
```bash
if [[ "$FIX_EXIT" -eq 0 ]]; then
  echo "$CHANGED_FILES" | xargs git add
fi
```

.githooks/pre-commit
```diff
@@ -0,0 +1,85 @@
...
+
+    # Restore unstaged changes
+    if [[ "$STASH_NEEDED" == true ]]; then
+      git stash pop --quiet
```
kadel: `git stash pop --quiet` can fail with a merge conflict if auto-fixes touch the same lines as unstaged changes. Under `set -e`, this kills the script immediately with a cryptic git error and the stash entry is not dropped (pop failure preserves it).

If a `trap cleanup EXIT` is added (see comment above), this is automatically handled. Otherwise, guard it explicitly:
```bash
if [[ "$STASH_NEEDED" == true ]]; then
  if ! git stash pop --quiet; then
    echo "" >&2
    echo "WARNING: Could not cleanly restore unstaged changes." >&2
    echo "Run 'git stash pop' and resolve any conflicts." >&2
  fi
fi
```

scripts/e2e-code-quality.sh
```diff
@@ -0,0 +1,133 @@
...
+  shift
+fi
+
+# Print error message to stderr. Uses GitHub Actions annotation syntax
+# in CI, plain text locally.
+report_error() {
+  local msg="$1"
+  if [[ -n "${CI:-}" ]]; then
+    echo "::error::${msg}" >&2
+  else
```
kadel, line 34: GitHub Actions workflow commands (`::error::`, `::warning::`, etc.) are parsed from **stdout**. Writing them to stderr (`>&2`) means the annotation won't render in the Actions UI — this is a regression from the original inline workflow code which used plain `echo "::error::..."` (stdout).

```suggestion
    echo "::error::${msg}"
```

zdrapela:
Thank you for the review @kadel! Good points, I incorporated them and an additional fix for the trap. PR is now ready

kadel (approved):

PR #2431, opened 2026-05-07
chore: migrate plugin catalog generation

Zaperex:
### Description

Migrates the plugin catalog generation scripts (with some modifications) to the overlays repository. Changes the source of truth for the productized catalog index to the `default.packages.yaml` file (anything not in that file goes into a community catalog index (use -e to exclude the DPY file)).

Also moved the `default.packages.yaml` file into the overlays repo.

Fixes [RHIDP-13209](https://redhat.atlassian.net/browse/RHIDP-13209)


nickboldt (approved):
let's merge it and see what happens!

nickboldt:
some concerns from the bots:

image

PR #2445, opened 2026-05-11
fix(tests): run smoke tests with partial metadata

rostalan:
## Summary
- allow smoke tests to run when at least one published plugin has runnable metadata instead of requiring metadata for every export
- report published plugins skipped for missing metadata in workflow comments (missing .env behaviour unchanged)
- update smoke-test docs to describe the skip/fail behavior

## Test plan
- [x] Parse `.github/workflows/workspace-tests.yaml` with Python YAML
- [x] Verify partial-metadata scenario produces runnable plugins instead of skipping entirely
- [x] Verify no-metadata scenario still skips smoke tests
- [x] Verify missing required `smoke-tests/test.env` still triggers the skip path
- [x] Confirm the missing-metadata summary is formatted for GitHub comments

### DIFF
--- a/.github/workflows/auto-publish-pr.yaml
+++ b/.github/workflows/auto-publish-pr.yaml
@@ -206,6 +206,8 @@ jobs:
       metadata-validation-passed: ${{ steps.validate-metadata.outputs.validation-passed }}
       metadata-validation-errors: ${{ steps.validate-metadata.outputs.validation-errors }}
       metadata-validation-error-count: ${{ steps.validate-metadata.outputs.validation-error-count }}
+      metadata-warning-packages: ${{ steps.extract-metadata-warnings.outputs.metadata-warning-packages }}
+      metadata-warning-count: ${{ steps.extract-metadata-warnings.outputs.metadata-warning-count }}
 
     steps:
       - name: Checkout overlay repository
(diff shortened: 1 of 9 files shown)


rostalan:
tested here: github.com/rostalan/rhdh-plugin-export-overlays/pull/133

gashcrumb (approved):

nickboldt:
[12:26 PM]Nick Boldt would it be better to prevent merging changes that cause incomplete metadata to be included? Shouldn't this be a failure that causes the plugin owner to go back and FIX the problem rather than hiding it for a future person to discover?
[12:26 PM]Nick Boldt   instead of requiring metadata for every exportdon't all plugins need working metadata?
[12:27 PM]Stan Lewis  I guess if a plugin doesn't have metadata it's not going to be included in any catalog, would this break the catalog build?
[12:33 PM]Nick Boldt  currently we include all the metadata but once Frank's stuff is ready, there will be the "only RHDH core plugins (default.packages.yaml + supported plugins.txt)" catalog and a second one for "everything else"

Eventually we might have 3 catalogs so there's one for DP content in an incubator catalog
[12:34 PM]Nick Boldt  catalog "build" is more like "copy the yaml files into a new location and then generate an index.json of all the [ghcr.io](http://ghcr.io/) links"
[12:34 PM]Stan Lewis  I wonder, if we want to make metadata required (or it should be required), should that be a specific metadata check vs relying on this smoke test check. (edited) 
[12:36 PM]Stan Lewis  which I think is done as part of publish currently?
[1:03 PM]Nick Boldt  I would hope that you can't add a plugin without metadata... otherwise how does the extensions UI know the plugin exists? And yes, this should be a PR check, not a smoke check
[1:52 PM]Stan Lewis  probably be good to add these concerns to the PR.  I guess the question is, do these changes allow for running the smoketest check while the metadata files aren't ready yet to provide for iterating on this easily in a PR before it's ready for merging.  Looking at some past merged PRs, the publish check includes the metadata validation, I don't think this PR gets around that.

nickboldt (changes requested):
I would prefer if we don't allow partial metadata. Instead we should prevent adding plugins which don't include their associated metadata. 

Unless there's a use case for plugins being included in this repo for export, but NOT being in the extensions catalog for marketing them?

rostalan:
waiting for https://github.com/redhat-developer/rhdh-plugin-export-utils/pull/148.

PR #2449, opened 2026-05-11
chore: update orchestrator metadata to point to ghcr.io images

Zaperex:
### Description
Update orchestrator metadata to point to the ghcr.io images

### DIFF
--- a/workspaces/orchestrator/metadata/rhdh-bsp-orchestrator-backend-mod-loki.yaml
+++ b/workspaces/orchestrator/metadata/rhdh-bsp-orchestrator-backend-mod-loki.yaml
@@ -17,8 +17,7 @@ metadata:
     - automation
 spec:
   packageName: "@red-hat-developer-hub/backstage-plugin-orchestrator-backend-module-loki"
-  # Tag: 1.10.0--1.0.1, Build date: 2026-04-20T22:08:14Z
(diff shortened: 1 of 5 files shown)


nickboldt:
This might break e2e testing of helm/bundle. Need to confirm with @lholmquist before merging.

lholmquist:
@Zaperex why are you updating to use those images?

Zaperex:
@lholmquist mainly for consistency since overlays is more of a community repo so it should make more sense to have `ghcr.io` images. But if it's necessary, then the change isn't needed.

lholmquist:
It looks like at one point, those metadata files got updated to point to the `ghcr.io` images, https://github.com/redhat-developer/rhdh-plugin-export-overlays/commit/25bba47e35985af43882a11bf796185d5fc87b10. but moved back to quay images, https://github.com/redhat-developer/rhdh-plugin-export-overlays/commit/15b04ed2219daf5de337f4716eac0aeca81fb26e , then got updated to use the red hat registry here:
https://github.com/redhat-developer/rhdh-plugin-export-overlays/commit/df70e6c6f25aa3fe08cb5470da0b53a28e690000

So i'm not sure this change would break anything

subhashkhileri:
are we also planning to update other plugins that still have registry.access.redhat.com? like : 
 "@backstage/plugin-events-backend-module-github" and "@roadiehq/scaffolder-backend-module-http-request"

lholmquist:
after a rebase,  this is probably fine to have. based on my previous comment, https://github.com/redhat-developer/rhdh-plugin-export-overlays/pull/2449#issuecomment-4430941221 it doesn't look like this would break anything

nickboldt (approved):

PR #2456, opened 2026-05-12
Bump Matomo Analytics Plugin

mjholder:
This update's the matomo analytics plugin to use the newly released version 1.28.2.

This updated version of the plugin has been migrated to the new frontend system. The goal of this change is to fix a compatability issue with RHDH 1.9 where we were getting duplicate API registration when trying to configure the plugin.

JIRA: https://redhat.atlassian.net/browse/ENGPROD-9774


davidfestal (changes requested):
This PR should wait for the overlay to be upgraded to Backstage 1.50

PR #2471, opened 2026-05-15
chore: add backstage compatability override command

Zaperex:
### Description

Adds a `/override-backstage` command to generate a `backstage.json` file matching the `versions.json` file in the root directory 

Fixes [RHIDP-13378](https://redhat.atlassian.net/browse/RHIDP-13378)


.github/workflows/pr-actions.yaml
```diff
@@ -492,3 +493,141 @@ jobs:
...
+            const overrideResult = core.getInput('override_result');
+            const success = overrideResult === 'success';
+            const fork = core.getInput('fork') === 'true';
+
+            await github.rest.repos.createCommitStatus({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              sha: overlayCommit,
+              description: '${{ github.workflow }}',
+              state: success ? 'success' : 'failure',
```
gashcrumb: If the branch is from a fork does this potentially skip the checks?  I guess if there's a comment added to the PR informing the user there's still work to be done we maybe fail in this case?

.github/workflows/github-script/override-backstage-compatibility.js
```diff
@@ -0,0 +1,205 @@
...
+    const ociTagRegex = /(dynamicArtifact:\s+oci:\/\/[^:]+:)bs_[\d.]+__/g;
+    const supportedVersionsRegex = /(supportedVersions:\s+)[\d.]+/g;
+    if (!ociTagRegex.test(content) && !supportedVersionsRegex.test(content)) {
+      skippedCount++;
+      continue;
+    }
+
+    const newContent = content
+      .replace(
+        /(dynamicArtifact:\s+oci:\/\/[^:]+:)bs_[\d.]+__/g,
```
gashcrumb: I guess we have some cases [like this](https://github.com/redhat-developer/rhdh-plugin-export-overlays/blob/main/workspaces/scaffolder-backend-module-servicenow/metadata/backstage-community-plugin-scaffolder-backend-module-servicenow.yaml#L21) where the URL is in a string, does this cater for that?  I was never great at regex 😄

.github/workflows/pr-actions.yaml
gashcrumb: I guess if the PR touches multiple workspaces the scripts won't do a publish, so I guess green would make sense in this case.

polasudo (approved):

.github/workflows/github-script/override-backstage-compatibility.js
Zaperex: I just updated this to use the update-plugins-repo-refs workflow to update metadata so the metadata recomputing in that should handle this now

gashcrumb (approved):

PR #2481, opened 2026-05-18
fix(smoketest): fail early on unhealthy RHDH startup

rostalan:
Add a startup log gate after health checks so smoke tests fail on backend boot/config errors instead of only plugin load symptoms. This makes overlay smoke failures clearer and diagnostics easier.

### DIFF
--- a/.github/workflows/run-workspace-smoke-tests.yaml
+++ b/.github/workflows/run-workspace-smoke-tests.yaml
@@ -117,6 +117,27 @@ jobs:
           echo "RHDH did not become ready in time."
           exit 1
 
+      - name: Verify RHDH startup logs
+        run: |
+          set -euo pipefail
+          LOGS=$(docker logs rhdh 2>&1 || true)
+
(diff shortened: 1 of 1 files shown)


.github/workflows/run-workspace-smoke-tests.yaml
```diff
@@ -149,8 +171,8 @@ jobs:
               failures+=("$plugin")
             fi
           done
-          if echo "$LOGS" | grep -E "(InstallException|Error while adding OCI plugin|Failed to load dynamic plugin|dynamic plugin.*error)" >/dev/null; then
-            echo "Detected dynamic plugin loading errors in logs"
+          if echo "$LOGS" | grep -qiE "$SHARED_HARD_ERROR_PATTERN"; then
```
gashcrumb: When is this particular case hit?  Seems like normally the startup check would always catch this first, wouldn't it?

rostalan: I guess there could be an edge case, where the log is printed in between, but since the jobs run very fast, it makes little sense to check it twice in the span of <10 seconds. 
-> merged the two steps into the second one.

gashcrumb (approved):

PR #2501, opened 2026-05-21
fix(e2e): unpin rhdh version

jrichter1:
https://redhat.atlassian.net/browse/RHDHBUGS-3030 should be fixed, lets see if the latest CI image is going to work

### DIFF
--- a/run-e2e.sh
+++ b/run-e2e.sh
@@ -38,7 +38,7 @@ cd "$SCRIPT_DIR"
 # These use defaults that can be overridden via environment variables.
 
 # RHDH deployment
-# export RHDH_VERSION="${RHDH_VERSION:-1.10}"             # RHDH version to deploy (e.g., "1.10", "next")
+export RHDH_VERSION="${RHDH_VERSION:-1.10}"             # RHDH version to deploy (e.g., "1.10", "next")
 export INSTALLATION_METHOD="${INSTALLATION_METHOD:-helm}" # "helm" or "operator"
 
 # Playwright
--- a/run-e2e.sh
+++ b/run-e2e.sh
@@ -38,7 +38,7 @@ cd "$SCRIPT_DIR"
 # These use defaults that can be overridden via environment variables.
 
 # RHDH deployment
-export RHDH_VERSION="${RHDH_VERSION:-1.10}"             # RHDH version to deploy (e.g., "1.10", "next")
+export RHDH_VERSION="${RHDH_VERSION:-1.11}"             # RHDH version to deploy (e.g., "1.10", "next")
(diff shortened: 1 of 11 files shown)


run-e2e.sh
```diff
@@ -39,7 +39,6 @@ cd "$SCRIPT_DIR"
 
 # RHDH deployment
 # export RHDH_VERSION="${RHDH_VERSION:-1.10}"             # RHDH version to deploy (e.g., "1.10", "next")
```
subhashkhileri:
```suggestion
export RHDH_VERSION="${RHDH_VERSION:-1.10}"             # RHDH version to deploy (e.g., "1.10", "next")
```

sanketpathak:
The orchestrator tests failure will be fixed with my pr https://github.com/redhat-developer/rhdh-plugin-export-overlays/pull/2473

run-e2e.sh
jrichter1: even if the same thing is set in the job itself?

nickboldt: should this move to 1.11 until we can move everything over to 2.1?

subhashkhileri: > even if the same thing is set in the job itself?

its there for reference, so when someone looks at the script they have idea what has to be changed to use different version.

subhashkhileri: @nickboldt there is no chart for 1.11 yet : https://quay.io/repository/rhdh/chart?tab=tags

nickboldt: There is now: check for `next-1.11`  or `1.11-8-CI`

run-e2e.sh
```diff
@@ -38,8 +38,7 @@ cd "$SCRIPT_DIR"
 # These use defaults that can be overridden via environment variables.
 
 # RHDH deployment
-# export RHDH_VERSION="${RHDH_VERSION:-1.10}"             # RHDH version to deploy (e.g., "1.10", "next")
-export RHDH_VERSION="1.10-114-CI"                       # Pinned due to RHDHBUGS-3030
+export RHDH_VERSION="${RHDH_VERSION:-1.10}"             # RHDH version to deploy (e.g., "1.10", "next")
```
nickboldt: for main this should be 1.11 for now, and 2.1 once we switch everything to that version. Right?

nickboldt:
```suggestion
export RHDH_VERSION="${RHDH_VERSION:-1.11}"             # RHDH version to deploy (e.g., "1.10", "next")
```

Perhaps you could read this value from https://github.com/redhat-developer/rhdh/blob/main/package.json#L3 (or the correct release-* branch) so it doesn't need updating when we branch or bump to a new version?

subhashkhileri (approved):
/lgtm

subhashkhileri (approved):
/lgtm

PR #2516, opened 2026-05-26
Pin `es5-ext` to version `0.10.64` in backstage monorepo

davidfestal:
Attempt to fix https://redhat.atlassian.net/browse/RHIDP-14012


kim-tsao (approved):
/lgtm

PR #2519, opened 2026-05-26
add gitlab-events to default.packages.yaml

04kash:
Add gitlab events to default.packages.yaml


nickboldt:
This is only required / allowed if Ben wants this new package to be "part of the CORE RHDH experience"; if not it can still be a discoverable plugin via the Extensions UI, but won't have its default config available to `{{inherit}}` from when extending the DPDY file for a specific install / configuration.

04kash:
> This is only required / allowed if Ben wants this new package to be "part of the CORE RHDH experience"; if not it can still be a discoverable plugin via the Extensions UI, but won't have its default config available to {{inherit}} from when extending the DPDY file for a specific install / configuration.

cc: @benwilcock Are you good with this?

The github events plugin is also included so I think it would make sense for the gitlab one to be included too

PR #2525, opened 2026-05-27
fix(homepage): update homepage frontend and backend

christoph-jerolimov:
Update homepage plugin to https://github.com/redhat-developer/rhdh-plugins/pull/3240

# Releases

### Patch Changes

-   ab8323b: Fix issue that defaul widget props are ignored when customization is enabled.

### Patch Changes

-   ab8323b: Fix conditional permission checks.


karthikjeeyar (changes requested):
`appConfigExamples` needs update (`ref` is missing there) in homepage-backend yaml file.

PR #2547, opened 2026-06-02
chore: add tag fallback to use outdated tags for best effort catalog index

Zaperex:
### Description

Syncs downstream changes by adding a fallback mechanism to attempt to use an existing older tag for an image

Plugins with missing builds are now excluded from catalog index, but catalog index will still build to prevent one broken plugin from blocking the entire catalog index build.

All details on failing/outdated plugins will be displayed in the wiki.

Fixes [RHIDP-13209](https://redhat.atlassian.net/browse/RHIDP-13209)

### DIFF
--- a/scripts/generatePluginBuildInfo.py
+++ b/scripts/generatePluginBuildInfo.py
@@ -15,13 +15,14 @@
 # - midstream: from container env MIDSTREAM_REPO
 
 import argparse
+import hashlib
 import json
 import os
+import re
 import sys
 from pathlib import Path
 import requests
 import yaml
-import hashlib
 
 from plugin_utils import (
     BuildReport,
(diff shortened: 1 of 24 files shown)


scripts/generatePluginBuildInfo.py
```diff
@@ -116,12 +115,104 @@ def get_output_registry_reference(registry_reference: str) -> str:
...
+            try:
+                parts.append((0, int(p)))
+            except ValueError:
+                parts.append((1, p))
+        return parts
+
+    return sorted(matched, key=version_key)
+
+
+def resolve_fallback_tag(registry_reference: str) -> str | None:
```
nickboldt, line 248: would love to see some tests or expanded logic for what we mean by `latest published tag with the same prefix` 

for example if we're working on 2.1.1 will this fall back to 1.10? or only to 2.1.0?

scripts/generatePluginBuildInfo.py
```diff
@@ -116,12 +115,104 @@ def get_output_registry_reference(registry_reference: str) -> str:
...
+        repository = name_part.rsplit(':', 1)[0] if ':' in name_part else name_part
+        tag = image_and_tag.split('@', 1)[1]
+    elif ':' in image_and_tag:
+        repository, tag = image_and_tag.rsplit(':', 1)
+    else:
+        return None
+    return registry, repository, tag
+
+
+def list_tags_with_prefix(registry: str, repository: str, prefix: str, auth, headers: dict) -> list[str]:
```
nickboldt, line 168: can we see some sample output / test data here so we know what is returned? asking because Konflux builds have A LOT of garbage tags, and we only want to find those that look like `x.y.z--a.b.c` or `x.y--a.b.c`... none of the other intermediate tags for sbom, signing, attribution, pull requests, push jobs,etc.

scripts/generatePluginBuildInfo.py
Zaperex: No it won't allow falling back to 1.10 (only same RHDH version). But sure having some tests for these scripts would be nice for sure. I'll add some.

PR #2563, opened 2026-06-04
chore(backstage): export some additional plugins

gashcrumb:
This change adds entries to produce dynamic plugins for some modules that are currently statically loaded by RHDH to eventually enable them to be loaded dynamically.

Assisted-By: Cursor Desktop

rh-pre-commit.version: 2.3.2


gashcrumb:
looks like it'd be helpful to also have an RHDH image that doesn't have these plugins included, using [this PR](https://github.com/redhat-developer/rhdh/pull/4922) for the moment to get one of those.

gashcrumb:
think I will go about this a bit differently given the scaffolder native module dependency.

PR #2577, opened 2026-06-09
feat: update DPDY script to emit enabled instead of disabled

hopehadfield:
Updates the DPDY generation script to instead use `enabled` as opposed to `disabled`

Works in conjunction with https://github.com/redhat-developer/rhdh/pull/4937

[RHIDP-11988](https://redhat.atlassian.net/browse/RHIDP-11988)


Zaperex (changes requested):
Please also update https://github.com/redhat-developer/rhdh-plugin-export-overlays/blob/fde86c4d3298de271f7a29b206ccaad974dd91ae/scripts/generateCatalogIndex.py#L132-L134 to use `enabled`

Zaperex (commented):
Just a few things that might need to also be updated:

- The e2e-tests and smoke tests still use the old `disabled` format as well.
  https://github.com/redhat-developer/rhdh-plugin-export-overlays/blob/b1fda1acc104f29107b3ba914a878bf28d74292f/.github/workflows/workspace-tests.yaml#L267

- Also I noticed that extensions plugin tests seem to set `disabled` too so we may need to also update https://github.com/redhat-developer/rhdh-plugins/tree/main/workspaces/extensions to use `enabled` instead?
https://github.com/redhat-developer/rhdh-plugin-export-overlays/blob/b1fda1acc104f29107b3ba914a878bf28d74292f/workspaces/extensions/e2e-tests/tests/specs/extensions.spec.ts#L247

- Then there are all the various `dynamic-plugins.yaml` configs in the overlays e2e tests for each workspace that are still using `disabled` which probably should be updated as well?
  https://github.com/redhat-developer/rhdh-plugin-export-overlays/blob/b1fda1acc104f29107b3ba914a878bf28d74292f/workspaces/homepage/e2e-tests/tests/config/dynamic-plugins.yaml#L3

- There's also a few `catalog-entities` fields that still include `disabled` in their documentation: https://github.com/redhat-developer/rhdh-plugin-export-overlays/blob/b1fda1acc104f29107b3ba914a878bf28d74292f/catalog-entities/extensions/plugins/lightspeed.yaml#L111

- Then the wiki page examples also need to be updated in https://github.com/redhat-developer/rhdh-plugin-export-overlays/tree/main/user-guide

We can handle these in a separate PR though so other than that, this LGTM.

Zaperex (approved):

PR #2579, opened 2026-06-09
chore(theme): pilot rhdh-plugins E2E coverage upload (e2e-test-utils 2.1.0)

gustavolira:
Pilots the **end-to-end coverage upload** for an `rhdh-plugins`-sourced workspace, now that the full chain is on main.

`theme` is a good pilot: sourced from `redhat-developer/rhdh-plugins` (the Codecov project we own, per the #2578 gate), a frontend-plugin (collectable browser coverage), and guest-auth (no Keycloak → reliable).

## The chain being validated
1. `__coverage` single-layer instrumented images + source-remapped lcov — #2574 (merged)
2. `_coverageCollector` fixture + `__coverage` image swap — `e2e-test-utils@2.1.0` (this bump)
3. `VAULT_CODECOV_TOKEN` fallback — #2576 (merged)
4. Upload only for the owned project — #2578 (merged)
5. `redhat-developer/rhdh-plugins` token in the Vault secret (updated)

## Expected on `/publish`
- `[remap] N source files` (real lcov)
- Codecov upload to **`redhat-developer/rhdh-plugins`**, flag `e2e-theme`, at theme's own source commit — where its files and commit actually exist (unlike the tech-radar pilot, which was correctly skipped as a community-plugins source).

tsc/lint/prettier pass; `yarn install --immutable` confirms the lockfile.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2580, opened 2026-06-09
feat: upload E2E coverage to this repo's Codecov project

gustavolira:
## Problem

`upload-coverage.sh` attributes E2E coverage to the upstream source repo + the historical `repo-ref` commit from `source.json`. Codecov finalizes a commit's report shortly after that commit's own CI completes, so uploads to an already-finalized historical commit are **accepted by the API but never displayed** — the commit page stays at *"No coverage data is available due to incomplete uploads... this page will not reflect the latest results"*. The data uploads and dies, regardless of org or token. (Observed on the #2579 pilot: theme's `repo-ref` points at a ~29-day-old rhdh-plugins commit already finalized with 21 uploads from its own CI.)

## Solution

Upload to **this overlay repo's own Codecov project**, attributed to the **fresh overlay commit** being tested:

- `--slug redhat-developer/rhdh-plugin-export-overlays` (override via `CODECOV_UPLOAD_SLUG`)
- `--sha` resolution: `PULL_PULL_SHA` (Prow presubmits — the checkout is a synthetic merge GitHub doesn't know, so use the PR head) → `GITHUB_SHA` (Actions) → `git rev-parse HEAD` (periodics/local)
- `--pr` from `PULL_NUMBER`/`GIT_PR_NUMBER` when available
- Flag stays `e2e-<workspace>`

This also removes the need for other orgs' tokens and keeps RHDH-specific E2E numbers off upstream community dashboards (a concern raised by Codecov when discussing cross-repo attribution).

With all workspaces reporting to a single dashboard, the rhdh-plugins-only gate from #2578 no longer makes sense — dropped, along with the `source.json` read and the `git ls-remote` ref resolution (one less network call).

## codecov.yml (new)

- `flag_management.default_rules.carryforward: true` — each PR exercises a single workspace's suite; other workspaces' flags carry forward from the last commit that measured them instead of zeroing out
- `coverage.status: project/patch off` + `require_ci_to_pass: false` — coverage is informational, never gates a PR
- `comment.layout: flags` — per-flag table on PRs

Validated against `https://codecov.io/validate` → `Valid!`

## Trade-off

This repo doesn't contain the plugin sources, so Codecov can't render line-level annotations — only the per-flag percentage and trend, which is the metric we want from E2E runs.

## Validation plan

1. The Codecov project for this repo already exists (`active: false` — first valid upload activates it)
2. ⚠️ `VAULT_CODECOV_TOKEN` in Vault was added for the rhdh-plugins project — if it's repo-scoped rather than an org/global upload token, the upload will be rejected and this repo's own token needs to be added to the `test-credentials/rhdh-plugin-export-overlays` Vault secret
3. After merge, re-run `/test e2e-ocp-helm` on #2579 (theme pilot on e2e-test-utils 2.1.0) and confirm the `e2e-theme` flag shows live data at https://app.codecov.io/gh/redhat-developer/rhdh-plugin-export-overlays

Refs: #2574, #2576, #2578, #2579

### DIFF
--- a/scripts/upload-coverage.sh
+++ b/scripts/upload-coverage.sh
@@ -31,6 +31,8 @@
 #                         Default: redhat-developer/rhdh-plugin-export-overlays.
 #   PULL_PULL_SHA       - PR head SHA (set by Prow presubmits).
 #   PULL_NUMBER         - PR number (set by Prow presubmits).
+#   PULL_BASE_REF       - Base branch (set by Prow postsubmits) — used as the
+#                         upload branch when there is no PR number.
 #   GITHUB_SHA          - Commit SHA (set by GitHub Actions).
 #   GIT_PR_NUMBER       - PR number fallback (exported by the E2E CI step).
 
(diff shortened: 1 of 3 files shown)


scripts/upload-coverage.sh
kadel, line 47: this variable is not used anymore, it can be removed

PR #2582, opened 2026-06-10
fix: clarify publish comment sections for failures and images

rostalan:
Issue seen here:
image
Separated failed plugin exports from published image section and added checkmark for visible separation.


rostalan:
After: 
image

gashcrumb (commented):
👍

PR #2583, opened 2026-06-10
chore: don't pin ghcr digest in unit tests

Zaperex:
### Description
Update catalog index unit tests to not expect a fixed digest from GHCR since we're constantly overwriting the tags with each push to main.


hopehadfield (approved):
Thanks for the fix!

PR #2636, opened 2026-06-19
chore: revert catalog-entities filtering

Zaperex:
### Description
Reverts the `catalog-entities` filtering back to the old logic where everything in `catalog-entities` is included in each `plugin-catalog-index`.

Fixes https://redhat.atlassian.net/browse/RHIDP-14969 

Test Commit for catalog index regen with filtering removed: https://github.com/Zaperex/rhdh-plugin-export-overlays/commit/8864de6c0aae488ad7069261aa05d15a675bf505#diff-191d2ed9aa2d4e3d45519d76d1b2e50d1f482dc8bb583d67f7d5751e23d0b247L1085

### DIFF
--- a/scripts/generateCatalogIndex.py
+++ b/scripts/generateCatalogIndex.py
@@ -31,16 +31,10 @@ from plugin_utils import (
     log_warn,
     log_error,
     set_debug,
-    load_and_resolve_to_stems,
 )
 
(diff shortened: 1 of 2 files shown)


hopehadfield (approved):
/lgtm

PR #2643, opened 2026-06-22
Moving status from comment to attr in default.packages.yaml

alizard0:
It moves the **plugin status** (`generally-available` and `tech-preview`) to the _default.packages.yaml_ from inline comment to attribute.

### DIFF
--- a/default.packages.yaml
+++ b/default.packages.yaml
@@ -2,68 +2,112 @@
 #
 # Do not add packages to this list without prior agreement from PM
 #
-# New packages should only be "tech-preview" or "generally-available"
+# New packages *status* should only be "tech-preview" or "generally-available"
 #    16 are tech-preview, and should be promoted via a PM-sponsored Feature to generally-available at some point
 #    06 are community, and should be removed over time
 packages:
-  enabled: # should onyly include GA (or TP with a path to GA) content here
-  - package: '@backstage-community/plugin-analytics-provider-segment' # generally-available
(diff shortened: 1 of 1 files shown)


Zaperex (approved):
/lgtm

nickboldt:
had not considered that the `support level` would be more than a comment (eg., used as something APIish or queryable) but if that's the direction we're going, cool. +1

default.packages.yaml
```diff
@@ -2,68 +2,120 @@
...
+  - package: '@backstage-community/plugin-analytics-provider-segment'
+    status: "generally-available"
+  - package: '@backstage-community/plugin-scaffolder-backend-module-regex'
+    status: "generally-available"
+  - package: '@backstage/plugin-techdocs'
+    status: "generally-available"
+  - package: '@backstage/plugin-techdocs-backend'
+    status: "generally-available"
+  - package: '@backstage/plugin-techdocs-module-addons-contrib'
+    status: "generally-available"
```
nickboldt, line 19: this should really be `support`, not `status` but hey. Why be consistent with what's in the catalog entity fields?

nickboldt: https://github.com/redhat-developer/rhdh-plugin-export-overlays/blob/main/catalog-entities/extensions/plugins/analytics-provider-segment.yaml#L33 (`.spec.support.level`)

https://github.com/redhat-developer/rhdh-plugin-export-overlays/blob/main/workspaces/acr/metadata/backstage-community-plugin-acr.yaml#L25 (`.spec.support`)

PR #2661, opened 2026-06-25
Test: bulk import scaffolder template tests

dom-aug:
Adds scaffolder template tests as per https://redhat.atlassian.net/browse/RHIDP-8402


lholmquist:
@dom-aug  One of those test failures  is related to this major change, https://github.com/redhat-developer/rhdh-plugins/pull/4349,  I think this is the test, https://github.com/redhat-developer/rhdh-plugin-export-overlays/blob/main/workspaces/bulk-import/e2e-tests/tests/specs/bulk-import-orchestrator.spec.ts#L93.

basically, the ability to use a PAT to import using the orchestartor flow was removed.  You now have to use a github App for it to work.  I wonder if we should just comment out that test for now.  Unless the setup for a github app is already part of the e2e tests somewhere.  

There is a error banner that is displayed, so maybe that test could just test the existence of that banner when using the PAT?

dom-aug:
@lholmquist  Thanks for bringing that up. Yes, that might be the reason you point out. The other failed test was caused by bad timing of GH 2FA in the test. I have seen it before and rerunning the tests was enough.

I have changed the test to expect the error since that can be the expected behavior without having Github App, from a certain point of view.

lholmquist (approved):

PR #2714, opened 2026-06-26
feat: Docker-free in-process smoke harness for backend dynamic plugins

gustavolira:
## What

A Docker-free, in-process smoke harness (`smoke-tests-native/`) that validates RHDH
dynamic **backend** plugins: it installs them with the published
`@red-hat-developer-hub/cli-module-install-dynamic-plugins` CLI, then boots them with
`startTestBackend` (`@backstage/backend-test-utils`) — no Docker container, no cluster.
About 20x faster than the per-workspace `docker run rhdh` smoke-test for that scope.

The loader / module-resolution / plugin-config are ported from RHDH's already-green
nightly test (`plugin-dynamic-loading.spec.ts`, redhat-developer/rhdh#4967).

## How it runs in CI

A dedicated, separate workflow `.github/workflows/native-smoke.yaml` — independent of the
Docker-based `run-workspace-smoke-tests.yaml`:

- `pull_request` (paths `smoke-tests-native/**`): validates the harness on every change here.
- `workflow_dispatch`: run against any plugin ref on demand.

It installs skopeo, builds with esbuild, runs `yarn smoke`, and uploads `results.json`.
There is no `docker run` anywhere. The PR's run is green — the harness installed the
plugin and booted it in-process (`status: pass`).

## Scope (honest boundary)

- Covers: backend dynamic plugins that don't extend the catalog (e.g. scaffolder modules).
- Not yet: catalog-extending modules (catalog-backend doesn't boot standalone in the
  minimal harness — tracked as follow-up) and frontend plugins (backend-only; UI render is
  the NFS / app-next path). Those stay on the Docker smoke for now.

This is additive — it removes no existing Docker smoke-test. Migrating covered workspaces
off Docker is a separate, parity-gated step.

## Verification

- CI green: SonarCloud + the new `smoke` check.
- Local (Node 24): `tsc --noEmit` clean; `yarn install --immutable` passes; harness run =
  `pass` on a good plugin, `error` on a bad ref.
- Benchmark vs Docker (same plugin, real images): ~5s vs ~104s.

Tickets: RHIDP-15075, RHIDP-15076, [RHIDP-13530](https://redhat.atlassian.net/browse/RHIDP-13530) (epic [RHIDP-13501](https://redhat.atlassian.net/browse/RHIDP-13501)).


rostalan (commented):
In my opinion, this is a step in the good direction, I am just worried that any potential gain atm is negated by the lack of coverage of the frontend and catalog plugins. If I still need to run the old smoke tests on the workspace to test these plugins, what good is it for to have the backend tested faster?
Edit: did not notice we had pure backend workspaces, it makes sense to run it in that case. It could also catch backend issues faster.

smoke-tests-native/src/native-smoke.ts
```diff
@@ -0,0 +1,247 @@
...
+
+    const manifest = discoverPlugins(root);
+    console.log(
+      `▶ manifest: ${manifest.backend.length} backend, ${manifest.frontend.length} frontend`,
+    );
+
+    // Let extracted plugins (under a temp dir) resolve their @backstage/* peers here.
+    patchModuleResolution(HARNESS_NODE_MODULES);
+    const backendPlugins = manifest.backend.filter(
+      (p) => !KNOWN_FAILURES.has(p.dirName),
```
rostalan: Shouldn't the workflow at least print a warning that plugins were skipped? 
If a workspace has only one backend plugin, which is listed in KNOWN_FAILURES, it would get skipped and it the validation would succeed for 0 plugins?

smoke-tests-native/src/plugin-config.ts
```diff
@@ -0,0 +1,70 @@
...
+ */
+
+/**
+ * Known failures + startup config overrides (ported from RHDH PR #4967:
+ * e2e-tests/playwright/utils/plugin-config.ts). Some backend plugins/modules
+ * validate config at boot, so startTestBackend needs a root config with (dummy)
+ * values for them; others can't load in a test env and are skipped.
+ */
+
+import type { JsonObject } from "@backstage/types";
```
rostalan, line 18: should this be in package.json?

smoke-tests-native/src/native-smoke.ts
```diff
@@ -0,0 +1,247 @@
...
+
+  const out = values.out ?? "results.json";
+  const dynamicPlugins = values["dynamic-plugins"];
+
+  if (!dynamicPlugins) {
+    console.error("Provide --dynamic-plugins <dynamic-plugins.yaml>.");
+    return 2;
+  }
+
+  const cliVersion = run(process.execPath, [CLI_BIN, "--version"]);
```
rostalan: if this fails, no results.json is written. move to the try block?

rostalan (changes requested):

smoke-tests-native/src/loader.ts
```diff
@@ -0,0 +1,147 @@
...
+    const dir = join(root, entry.name);
+    const pkgPath = join(dir, "package.json");
+    if (!existsSync(pkgPath)) continue;
+
+    let pkg: { name?: string; version?: string; backstage?: { role?: string } };
+    try {
+      pkg = JSON.parse(readFileSync(pkgPath, "utf8"));
+    } catch {
+      // A malformed package.json in one dir shouldn't abort discovery of the rest.
+      continue;
```
rostalan: so if the package.json is malformed, the plugin validation is skipped with no warning/fail ?

smoke-tests-native/src/native-smoke.ts
```diff
@@ -0,0 +1,247 @@
...
+    options: {
+      "dynamic-plugins": { type: "string" },
+      out: { type: "string" },
+    },
+  });
+
+  const out = values.out ?? "results.json";
+  const dynamicPlugins = values["dynamic-plugins"];
+
+  if (!dynamicPlugins) {
```
rostalan: this only checks if the argument is passed, not that the file actually exists.

smoke-tests-native/src/native-smoke.ts
gustavolira: Good catch. It now warns when backend plugins are filtered by `KNOWN_FAILURES`, plus a distinct "nothing validated" warning when 0 plugins end up validated (so a known-failure-only workspace can't pass silently). The skipped list is included in `results.json` too.

smoke-tests-native/src/native-smoke.ts
gustavolira: Fixed — moved the CLI version check and temp-dir setup inside the `try`, so a failure there writes a `results.json` (`status: error`) instead of exiting with no report.

smoke-tests-native/src/loader.ts
gustavolira: Right — it now warns loudly (with the path) instead of skipping silently, while still not aborting discovery of the other plugin dirs.

smoke-tests-native/src/native-smoke.ts
gustavolira: Added an `existsSync` check with a clear "file not found" message, so it fails fast with a useful error instead of a generic ENOENT during the copy.

smoke-tests-native/src/plugin-config.ts
gustavolira: Considered it — kept it in TS: it's ported as-is from RHDH's `plugin-config.ts` (stays in sync with their list) and stays type-checked. Added a comment noting these lists apply to multi-plugin/whole-workspace runs. Happy to externalize to `package.json` if you'd prefer that for per-workspace overrides later.

gustavolira:
Heads-up that the frontend / UI-render half this PR scopes out ("backend-only; UI render is the NFS / app-next path") now has a proven, Docker-free path in redhat-developer/rhdh#5005.

That PR is basically the frontend complement to this one. It boots the RHDH backend + the legacy app (`packages/app`) dev servers in-process — no cluster, no image build — and runs an existing Playwright spec against them. Dynamic frontend plugins are installed exactly the way you do here: the published `install-dynamic-plugins` CLI pulling from OCI via skopeo, so no `dynamic-plugins/dist` source build. It's green in CI in ~4.5 min — the `dynamic-home-page` plugin (from this repo's OCI registry) renders "Welcome back!" + Quick Access off-cluster and the existing guest-signin spec passes unmodified.

Two things from it that are directly relevant here:

- The `install-dynamic-plugins` + OCI populate works for **frontend** plugins too, not just backend — the same step extends to UI render.
- Render nuance worth knowing: plugins that mount on a **route** (e.g. the home page at `/`) render with just the plugin + the repo's `app-config.dynamic-plugins.yaml`; plugins that mount **app-level** (e.g. `global-header`) load fine (scalprum serves them) but need extra config to actually mount in the layout.

Honest gap for doing the same in this repo: the frontend harness needs the full RHDH app to boot, which the overlays repo doesn't have locally — so a "frontend smoke" here wouldn't be a trivial port of this backend harness. But the approach and the OCI-for-frontend validation are proven if we want to add the frontend counterpart down the line. Together they'd cover the whole Docker-free/cluster-free smoke: backend loads (this PR) + frontend renders (#5005).

smoke-tests-native/src/loader.ts
```diff
@@ -0,0 +1,154 @@
...
+}
+
+/**
+ * Check that a frontend plugin's bundle artifacts EXIST (scalprum/remoteEntry). This is a
+ * presence check only — the bundle is never loaded or evaluated.
+ */
+export function validateFrontendBundle(plugin: PluginEntry): string | null {
+  const has = (rel: string) => existsSync(join(plugin.path, rel));
+  if (!has("package.json")) return "missing package.json";
+  if (!has("dist-scalprum") && !has("dist/remoteEntry.js")) {
```
gashcrumb: It'd also be good to cater for new frontend system plugins at this point as well, as once we make the switch this will need to be updated.  There's a few examples available, [app-auth](https://github.com/redhat-developer/rhdh-plugin-export-overlays/pkgs/container/rhdh-plugin-export-overlays%2Fred-hat-developer-hub-backstage-plugin-app-auth) is a simple one that *only* supports the new frontend system, and probably [tech-radar](https://github.com/redhat-developer/rhdh-plugin-export-overlays/pkgs/container/rhdh-plugin-export-overlays%2Fbackstage-community-plugin-tech-radar) is a good example of a frontend plugin that currently supports both frontend systems.

smoke-tests-native/src/plugin-config.ts
```diff
@@ -0,0 +1,74 @@
...
+ *
+ * Licensed under the Apache License, Version 2.0.
+ */
+
+/**
+ * Known failures + startup config overrides (ported from RHDH PR #4967:
+ * e2e-tests/playwright/utils/plugin-config.ts). Some backend plugins/modules
+ * validate config at boot, so startTestBackend needs a root config with (dummy)
+ * values for them; others can't load in a test env and are skipped.
+ *
```
gashcrumb, line 12: Does this still support workspaces that supply a smoke test [app-config.test.yaml](https://github.com/redhat-developer/rhdh-plugin-export-overlays/blob/7578b65121c667954a28424c4385281b2f8d95e8/workspaces/mcp-chat/smoke-tests/app-config.test.yaml#L4) and [test.env](https://github.com/redhat-developer/rhdh-plugin-export-overlays/blob/7578b65121c667954a28424c4385281b2f8d95e8/workspaces/acs/smoke-tests/test.env#L3) files?

smoke-tests-native/src/loader.ts
gustavolira: The check now covers both systems — `dist-scalprum/` + `plugin-manifest.json` (legacy) and `dist/remoteEntry.js` + `dist/mf-manifest.json` (new FE) — and results.json records which one(s) each plugin ships, so we can track the migration later.

Tested with both your examples, app-auth and tech-radar pass. Fun fact: the app-auth artifact actually ships both layouts (the export still generates the scalprum wrapper), so it reports `legacy + new-frontend-system` too.

smoke-tests-native/src/plugin-config.ts
gustavolira: It didn't — good catch. Added in 00ac3354: `--app-config` (extra config layer, `${VAR}`/`${VAR:-default}` resolved from the env) and `--test-env` (same as `docker run --env-file`; already-set vars win, so CI secrets beat the committed placeholders). Tested with mcp-chat's app-config.test.yaml and acs' test.env.

Named it `--test-env` because Node grabs `--env-file` for itself even after the script path. The workspace-mode follow-up PR picks these up automatically from `workspaces/<name>/smoke-tests/`.

gashcrumb:
looking good with the recent updates, probably worth tackling those two sonarcloud issues and then I'm :+1:

gustavolira:
> looking good with the recent updates, probably worth tackling those two sonarcloud issues and then I'm 👍

Thank you very much @gashcrumb ! Sonar issues fixed

PR #2715, opened 2026-06-26
chore(backstage): fix bitbucket cloud export

gashcrumb:
This change ensures the bitbucket-cloud-common package is embedded in the catalog/scaffolder bitbucket cloud modules which should sort errors seen in the smoke test output.

rh-pre-commit.version: 2.3.2


gashcrumb:
I'll put this commit over on the other PR instead of this.

PR #2735, opened 2026-07-07
feat: informational native-vs-Docker smoke parity job

gustavolira:
## What

Phase 1 of the native-smoke migration ([RHIDP-13530](https://redhat.atlassian.net/browse/RHIDP-13530)): a new `native-smoke` job in
`run-workspace-smoke-tests.yaml` runs the Docker-free in-process harness
(`smoke-tests-native/`, merged in #2714/#2731) on the **same `smoke-test-artifacts`
the container job consumes** — `dynamic-plugins.test.yaml` plus the optional
`app-config.test.yaml` (the extra `--config` mount) and `test.env` (the
`--env-file`). Identical input, so the verdicts are directly comparable.

A `parity-report` job renders the comparison in the run summary:

| Harness | Verdict |
|---|---|
| Docker (container) | pass/fail |
| Native (in-process) | pass/fail |

## Why

Every `/smoketest` and publish run now produces a native-vs-Docker parity data
point for free. Once the covered workspaces show consistent agreement, phase 2
adds a per-workspace allowlist so they skip the ~104s container in favor of the
~5s native run — that's where the CI time is saved.

## Guardrails

- **Strictly informational**: `continue-on-error` at job and step level — this can
  never block a PR, and the workflow-level outputs are untouched (still mapped to
  `jobs.run`), so callers are unaffected.
- Native failures on catalog-extending workspaces are expected at this stage
  (upstream catalog-backend boot issue) and called out in the summary text.
- Native `results.json` (with `schemaVersion`) is uploaded as
  `native-smoke-parity-results` for offline analysis.

## Verification

- Workflow YAML parses; jobs: `run` (unchanged), `native-smoke`, `parity-report`.
- The exact step command was simulated locally (Node 24) against a crafted
  `artifacts/` dir (`dynamic-plugins.test.yaml` + acs `test.env` + mcp-chat
  `app-config.test.yaml`): both files applied, `status: pass`.
- Note: this file is `workflow_call`-only, so this PR's own checks don't exercise
  it — the first real run happens on the next `/smoketest` or publish after merge,
  which is safe by construction (informational).

Tickets: [RHIDP-13530](https://redhat.atlassian.net/browse/RHIDP-13530) (epic [RHIDP-13501](https://redhat.atlassian.net/browse/RHIDP-13501)).

### DIFF
--- a/.github/workflows/run-workspace-smoke-tests.yaml
+++ b/.github/workflows/run-workspace-smoke-tests.yaml
@@ -246,3 +246,124 @@ jobs:
       - name: Cleanup
         if: always()
         run: docker rm -f rhdh || true
+
+  # Runs the Docker-free, in-process harness (smoke-tests-native/) on the same
+  # artifact the container smoke consumes, and compares the two verdicts in the
+  # run summary. Informational only — continue-on-error keeps it from ever
+  # blocking a PR.
+  native-smoke:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    continue-on-error: true
+    permissions:
+      contents: read
+      packages: read
+    outputs:
+      verdict: ${{ steps.verdict.outputs.verdict }}
+    steps:
+      - name: Checkout (harness source)
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
+
+      - name: Download smoke test artifacts
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8.0.1
+        with:
+          name: smoke-test-artifacts
+          path: ./artifacts
+
+      - name: Enable Corepack
+        # Before setup-node so its yarn-cache detection resolves Yarn 4 via corepack.
+        run: corepack enable
+
+      - name: Setup Node.js
+        uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
+        with:
+          node-version: 24
+          cache: yarn
+          cache-dependency-path: smoke-tests-native/yarn.lock
+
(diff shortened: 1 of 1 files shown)


gustavolira:
Since this file is `workflow_call`-only and the PR checks can't exercise it, I validated it end-to-end on my fork with a throwaway dispatch workflow that fabricates the same `smoke-test-artifacts` bundle the real callers upload and then `uses:` this reusable workflow.

Green run with all four jobs: https://github.com/gustavolira/rhdh-plugin-export-overlays/actions/runs/28868518531

- `run` (Docker container): pass
- `native-smoke`: pass — `results-parity.json` uploaded (`schemaVersion: 1`, backend 1/1 loaded + booted)
- `parity-report`: rendered `agree (pass)` in the run summary

The earlier iterations also exercised the failure paths for real: with an artifact missing the root `smoke-tests/app-config.yaml`, the container backend listens but every core plugin crashes on `Missing required config value at 'backend.baseUrl'` (while the plugin itself loads fine) — the report classified that docker=fail/native=pass divergence correctly, and a native-job infra failure renders as `no-data` instead of a false divergence.

gustavolira:
Closing in favor of a simpler approach. The weakness of this design is the consumption story: the parity report lands in per-run summaries and 90-day artifacts — places nobody watches systematically, so the data wouldn't actually drive the migration decision.

Replacement (PR incoming): skip the informational phase and go straight to an allowlist with an overlap window — for the workspaces already validated against the native harness, run the native smoke *alongside* the container smoke as a real check on their bump PRs. The parity evidence then shows up exactly where maintainers already look (PR checks, side by side), and after a period of sustained agreement a trivial follow-up drops the container job for those workspaces.

The end-to-end validation done here carries over: https://github.com/gustavolira/rhdh-plugin-export-overlays/actions/runs/28868518531 (native + container agreeing on the same artifact, plus the failure-path classification).

PR #2754, opened 2026-07-10
chore(backstage): export app-visualizer plugin

gashcrumb:
This change adds the app-visualizer to the list of plugins to export so it can be removed from RHDH's NFS app.

Fixes [RHIDP-15413](https://redhat.atlassian.net/browse/RHIDP-15413)

rh-pre-commit.version: 2.3.2


gashcrumb:
seems the quickstart plugin is now loading via the OCI artifact instead of the wrappers (rhdh-e2e-test-utils disables the wrapper) in these past few runs and getting in the way.  I wonder though if it's worth holding off addressing this until we remove the wrappers which would probably greatly simplify how this is handled in rhdh-e2e-test-utils, @subhashkhileri what do you think?

gashcrumb:
ok, merging this despite the test failures as this plugin isn't loaded by the legacy frontend; test failures are unrelated.

PR #2787, opened 2026-07-15
chore: update loki to GA status in DPY

kim-tsao:
Loki has been GA since 1.10 (see [metadata](https://github.com/redhat-developer/rhdh-plugin-export-overlays/blob/main/workspaces/orchestrator/metadata/rhdh-bsp-orchestrator-backend-mod-loki.yaml#L26)).  We should update DPY to reflect this.  

Needs backport to 1.10


nickboldt:
cherrypick in https://github.com/redhat-developer/rhdh-plugin-export-overlays/pull/2790

PR #2788, opened 2026-07-15
refactor(skill): port e2e-failure-analysis scripts to TypeScript

subhashkhileri:
## Summary
- Replace `diagnostics.py` and `download-artifacts.py` with TypeScript equivalents
- Scripts run via `node --experimental-strip-types` (Node 22+) — no `npx`, `tsx`, or Python dependency needed
- Uses native `fetch`, `fs/promises`, stream pipelines, and a proper concurrency pool
- Output is byte-identical to the Python originals (cross-tested against a real Prow URL)
- Updated all references in `SKILL.md` from `python3` to `node --experimental-strip-types`

## Test plan
- [x] Cross-tested `download-artifacts.ts` against real Prow URL — same output path as Python
- [x] Cross-tested `diagnostics.ts` — `diff` shows zero differences vs Python output
- [x] Tested `--project` filter — identical output
- [x] Verified `node --experimental-strip-types` works in fullsend sandbox (Node v22.22.1)
- [x] Fresh download (cache cleared) produces correct artifacts

Assisted-by: Claude Code


.fullsend/skills/e2e-failure-analysis/scripts/download-artifacts.ts
```diff
@@ -0,0 +1,168 @@
...
+    process.stderr.write(`ERROR: Could not parse URL: ${process.argv[2]}\n`);
+    process.exit(1);
+  }
+
+  const base = path.join("node_modules", ".cache", "e2e-artifacts");
+  const cacheDir = path.join(base, info.pr || "nightly", info.job_id);
+  const container = "redhat-developer-rhdh-plugin-export-overlays-ocp-helm";
+  const artifactsDir = path.join(cacheDir, container, "artifacts");
+
+  await fs.rm(cacheDir, { recursive: true, force: true });
```
gashcrumb, line 140: This appears to delete this cache directory with each run, is it worth updating [this description](https://github.com/redhat-developer/rhdh-plugin-export-overlays/blob/main/.fullsend/skills/e2e-failure-analysis/SKILL.md?plain=1#L34-L37), or implement a similar approach as the python script?

subhashkhileri: I have updated the description, thanks

gashcrumb (approved):

PR #2795, opened 2026-07-16
fix(e2e): ensure last row cells are visible in installed package table before sorting in extensions tests

sanketpathak:


workspaces/extensions/e2e-tests/tests/support/extensions.ts
```diff
@@ -38,7 +42,7 @@ export class ExtensionsPage {
       })
       .click();
     await expect(
-      this.page.getByText(pluginTitle + " " + " by " + " Red Hat" + badgeText, {
+      this.page.getByText(pluginTitle + " " + " by " + fromAuthor + badgeText, {
```
jrichter1: how about using a template string

workspaces/extensions/e2e-tests/tests/support/extensions.ts
```diff
@@ -29,7 +29,11 @@ export class ExtensionsPage {
     this.uiHelper = uiHelper;
   }
 
-  async clickReadMoreByPluginTitle(pluginTitle: string, badgeText: string) {
+  async clickReadMoreByPluginTitle(
+    pluginTitle: string,
+    badgeText: string,
+    fromAuthor: string = " Red Hat",
```
jrichter1: why the space at the start, when at L45 there already is a space in the expected text?

PR #2803, opened 2026-07-17
chore(wiki): add license prereq

hopehadfield:
Per discussion in Jul 16 COPE office hours


04kash (commented):
Should we also add a repo level Lisence file like in rhdh-plugins?

hopehadfield:
> Should we also add a repo level Lisence file like in rhdh-plugins?

Yes, I'll be adding that as soon as https://github.com/proberaum/backstage-plugins/pull/254 is merged

04kash (approved):

PR #2810, opened 2026-07-17
chore: add workflow to regenerate catalog-entities

Zaperex:
### Description
Workflow to automate auto-regen the `all.yaml` of the `catalog-entities` directory so we won't need to worry about [RHDHBUGS-2757](https://redhat.atlassian.net/browse/RHDHBUGS-2757) in the future


.github/workflows/sync-extensions-index.yaml
```diff
@@ -0,0 +1,66 @@
+name: Sync extensions catalog index
+
+on:
+  push:
```
gashcrumb, line 4: Was post-merge-only intentional here? With this trigger, a PR can still land with a stale `all.yaml` and `main` stays wrong until this job regenerates and pushes successfully.

Did you consider (or deliberately skip) a `pull_request` check that fails when the regenerated index differs from what's committed? That would shift the fix into the PR itself rather than relying on a follow-up bot commit.

.github/workflows/generate-catalog-index.yaml
```diff
@@ -95,7 +95,7 @@ jobs:
       - name: Setup Python
         uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
         with:
-          python-version: '3.11'
+          python-version: '3.12'
```
gashcrumb, line 98: Is the Python 3.11 → 3.12 bump required for the extensions index sync, or is it a convenience upgrade bundled into this PR? If it's independent, would you mind splitting it out (or calling out why it needs to land with this change)?

scripts/updateExtensionsIndex.py
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env python3
+"""Regenerate catalog-entities/extensions/{plugins,packages}/all.yaml Location
```
gashcrumb, line 2: The docstring mentions regenerating `packages/all.yaml` as well, but today `catalog-entities/extensions/` only has `plugins/` (and `collections/`) — no `packages/` directory. Is that anticipating a future layout, or just reflecting what `regenerate_all_yaml_files()` already supports?

PR #2864, opened 2026-07-21
fix(e2e): add missing kube config for argocd & topology tests

jrichter1:
should make argocd and topology e2e tests run properly again


sanketpathak:
Topology and ArgoCD tests are passing without the missing Kubernetes service API implementation error
/lgtm

PR #2866, opened 2026-07-21
fix(coverage): guard instrumented chunks against RHDH zip-bomb detection (RHDHBUGS-3470)

gustavolira:
## Summary
- Fixes [RHDHBUGS-3470](https://redhat.atlassian.net/browse/RHDHBUGS-3470): `__coverage` OCI images trip RHDH's zip bomb detection because Istanbul inflates large vendor chunks ~25x (topology's 1.7 MB vendor chunk → 44.6 MB), exceeding `install-dynamic-plugins`' per-entry `MAX_ENTRY_SIZE` limit and aborting e2e before any test runs.
- Root cause refinement: the oversized chunks are **vendor code** (lodash, d3, MUI, react-topology). `remap-coverage.cjs` already drops `node_modules` entries from the final lcov, so instrumenting them yields zero useful coverage — while being exactly what trips the detector.
- Fix: add a size guard to `scripts/instrument-plugin.sh` — after `nyc instrument`, any file whose instrumented output exceeds `MAX_INSTRUMENTED_ENTRY_SIZE` (default 15 MB, safely below RHDH's limit) is restored to its original uninstrumented version. App-code chunks stay instrumented; no `MAX_ENTRY_SIZE` bump needed, detector stays intact.
- Reverts the mitigation: `E2E_COLLECT_COVERAGE` default back to `true` in `run-e2e.sh` (ticket acceptance criteria).
- Touches the topology spec (comment only) to trigger the PR e2e job on the affected workspace.

## Local verification (against the published topology image)
- `dist/static/4697.*.chunk.js`: instrumented 44,720,109 B → restored original 1,784,283 B
- `dist-scalprum/static/3069.*.chunk.js`: instrumented 44,119,337 B → restored original 1,796,962 B
- 431 JS files remain instrumented (206 dist + 225 dist-scalprum); `remoteEntry.js` still carries `__coverage__`
- Final image: single layer, largest entry 11.2 MB (a sourcemap) — nothing above 15 MB

## Test plan
- [ ] `/publish` to build `__coverage` images with the guard
- [ ] PR e2e (`e2e-ocp-helm`) on topology: no "Zip bomb detected" in `install-dynamic-plugins.log`, tests run, coverage artifacts non-empty

### DIFF
--- a/run-e2e.sh
+++ b/run-e2e.sh
@@ -73,8 +73,10 @@ E2E_NIGHTLY_MODE="${E2E_NIGHTLY_MODE:-false}"
 # skipped silently (no -coverage images exist).
 #
 # To disable (faster local dev): E2E_COLLECT_COVERAGE=false
-# Disabled by default due to zip bomb detection failures — see RHDHBUGS-3470
-export E2E_COLLECT_COVERAGE="${E2E_COLLECT_COVERAGE:-false}"
+# Zip bomb detection (RHDHBUGS-3470) is handled at image-build time:
+# instrument-plugin.sh restores the original file for any chunk whose
+# instrumented output would exceed RHDH's per-entry size limit.
+export E2E_COLLECT_COVERAGE="${E2E_COLLECT_COVERAGE:-true}"
 
 # Local e2e-test-utils: absolute path to use a local build instead of npm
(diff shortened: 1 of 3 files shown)


gashcrumb (approved):

gashcrumb:
Looks like this fix gives us a running RHDH, I think we can merge this as the test failure isn't related to this change.

PR #2974, opened 2026-07-28
Promote app-defaults plugins to tech-preview and enable OOTB (RHIDP-15482)

gashcrumb:
## Summary

- Bump `app-auth` and `app-integrations` Package metadata from `dev-preview` to `tech-preview`
- Add both packages to `rhdh-supported-packages.txt` (Supported Plugins / Konflux path)
- Enable both under `default.packages.yaml` `packages.enabled` so OOTB NFS auth works beyond guest

**Why tech-preview (not GA):** GA needs Konflux security scanning, which is not wired yet. TP is enough for DPDY inclusion per PM guidance.

**Why enabled while TP:** These plugins carry the auth/SignInPage and SCM integration setup extracted from the legacy frontend. Without them enabled by default, customers only get guest auth after the NFS hard cut-over ([RHDHPLAN-846](https://redhat.atlassian.net/browse/RHDHPLAN-846) / [RHIDP-15077](https://redhat.atlassian.net/browse/RHIDP-15077)). This is an intentional exception to the 2.1 “GA-only in `default.packages.yaml`” header; comments in that file document it. Promote to GA in a follow-up once security scans land.

Keeps delivery **dynamic** (not statically baked) so custom SignInPage / SCMAuth setups remain replaceable.

Jira: https://redhat.atlassian.net/browse/RHIDP-15482

## Test plan

- [ ] `/publish` on this PR
- [ ] `/smoketest` after publish
- [ ] Confirm app-defaults E2E still passes (auth + integrations on app-next)
- [ ] Reviewer ack of TP-in-`enabled:` exception for NFS OOTB auth
- [ ] After merge: midstream sync + Konflux onboarding (Pyxis → KRD → PipelineRuns) per [ONBOARD_KONFLUX](https://gitlab.cee.redhat.com/rhidp/rhdh-plugin-catalog/-/blob/rhdh-1-rhel-9/docs/ONBOARD_KONFLUX.adoc)

### DIFF
--- a/default.packages.yaml
+++ b/default.packages.yaml
@@ -7,6 +7,7 @@
 #
 # Package support level should only be "generally-available" as of 2.1
 #    Tech-preview entries are commented out until promoted to generally-available
+#    Exception: app-auth / app-integrations (RHIDP-15482) — TP but enabled OOTB for NFS auth
 #    Community plugins have been removed https://redhat.atlassian.net/browse/RHIDP-13262
 packages:
   # should only include GA (or TP with a path to GA) content here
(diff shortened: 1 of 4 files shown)


Zaperex (approved):
/lgtm

PR #2980, opened 2026-07-28
refactor(fullsend): split e2e-fix monolith into triage + scaffold coder

subhashkhileri:
## Summary

- Replace the monolithic e2e-fix agent (770-line agent + 660-line post-script + 245-line schema) with a focused **e2e-triage agent** that classifies nightly E2E failures and creates per-workspace GitHub issues
- The existing **scaffold coder** (`rhdh/agents/code.md` + `rhdh/harness/code.yaml`) picks up each issue via `ready-to-code` label and implements the fix — zero custom coder code needed
- Net reduction: **~1,800 lines** of custom code replaced by **~880 lines**, with the CI feedback loop (scaffold fix agent) available for free in the future

### Architecture change

```
Before (monolith):
  nightly fails → e2e-fix agent (analyze + classify + dedup + fix + commit)
                → post-e2e-fix.sh (validate + push + PR + JIRA)

After (split):
  nightly fails → e2e-triage agent (analyze + classify + dedup + emit issue directives)
                → post-e2e-triage.sh (create/comment issues, add labels)
                → scaffold coder picks up ready-to-code issues → fix → PR
```

### New files
| File | Purpose |
|------|---------|
| `rhdh/agents/e2e-triage.md` | Triage agent: analyze failures, classify, dedup via tracking lines, emit issue directives |
| `rhdh/scripts/post-e2e-triage.sh` | Execute issue directives, cycle `ready-to-code` label, post summary |
| `rhdh/schemas/e2e-triage-result.schema.json` | Flat schema — no mode discriminator, no groups |
| `rhdh/harness/e2e-triage.yaml` | Harness with `role: triage`, trigger on `e2e-triage` label |
| `rhdh/policies/e2e-triage.yaml` | Network policy without package registries (triage doesn't build) |

### Deleted files
| File | Lines |
|------|-------|
| `rhdh/agents/e2e-fix.md` | 639 |
| `rhdh/scripts/post-e2e-fix.sh` | 657 |
| `rhdh/schemas/e2e-fix-result.schema.json` | ~200 |
| `rhdh/harness/e2e-fix.yaml` | 79 |
| `rhdh/policies/e2e-fix.yaml` | 106 |

### Key design decisions

- **Phase 1 artifact download**: Artifacts are downloaded once upfront via `download-artifacts.ts` and diagnostics run across all projects. Subagents reuse these paths instead of each downloading independently.
- **Skill invocations**: Subagents invoke `/e2e-failure-analysis` and `/playwright-trace` skills directly. If either skill fails to invoke, the subagent exits immediately with an error — no proceeding without skills.
- **Evidence-based classification**: Subagents return root cause evidence and classification inputs (unique code path, component reuse, preventability) — not `fix_category` suggestions. Classification happens in Phase 3 where cross-workspace context is available.
- **Tighter `infra_flake` criteria**: Requires differential diagnosis — if the same infrastructure component worked for other tests in the run, the failure is `test_fix`, not `infra_flake`. `infra_flake` is reserved for failures where no test code change would help.
- **Dedup via tracking lines**: Issues include `fullsend-tracking: workspace=X root-cause=Y branch=Z` lines searchable with GitHub's `in:body` qualifier. `linked:pr` eliminates separate PR searches.
- **`cycle_ready_to_code`**: When an issue exists but its PR was closed, the post-script removes then re-adds `ready-to-code` to force a new `issues.labeled` webhook.
- **Umbrella issues**: When ≥3 workspaces share the same root cause, one umbrella issue is created with tracking lines for each workspace → one coder run → one PR.
- **Labels via API**: Issue creation uses `gh issue create` without `--label`, then adds labels via the labels REST API so `ready-to-code` fires a proper `issues.labeled` event for coder dispatch. `add_label` emits `::warning::` on failure for observability.
- **AGENTS.md conventions**: Added "E2E Nightly Fix Conventions" section so the scaffold coder knows allowed paths, the `test.skip` pattern, and verification steps.
- **Runner env**: `GITHUB_ISSUE_URL` added to `env.runner` so the post-script can post summary comments on the trigger issue.
- **Timeout**: 60 minutes (up from 30) to accommodate multi-workspace analysis with subagents.

## Test plan

- [ ] Run `fullsend run e2e-triage` locally with a known failed nightly URL
- [ ] Verify `agent-result.json` contains correct classifications and issue directives
- [ ] Verify post-script creates issues with correct tracking lines and labels
- [ ] Verify scaffold coder picks up a test issue with `ready-to-code` label
- [ ] Verify dedup: run triage again → should emit `comment` not `create`
- [ ] Verify `cycle_ready_to_code`: close the PR → re-run triage → coder re-triggers

Assisted-by: Claude Code


.fullsend/rhdh/agents/e2e-triage.md
```diff
@@ -0,0 +1,500 @@
+---
+name: e2e-triage
+description: >-
+  Analyze E2E nightly test failures, classify root causes per workspace,
+  search for existing issues (dedup), and emit structured issue directives.
```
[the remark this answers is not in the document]
subhashkhileri: The agent needs `gh api` for read-only operations (dedup search, issue lookup) — sandbox network policy enforces read-only access.

.fullsend/rhdh/scripts/post-e2e-triage.sh
```diff
@@ -0,0 +1,276 @@
...
+    && echo "${GITLEAKS_SHA256}  /tmp/gitleaks.tar.gz" | sha256sum -c --quiet \
+    && tar xzf /tmp/gitleaks.tar.gz -C "${HOME}/.local/bin" gitleaks \
+    && rm /tmp/gitleaks.tar.gz; then
+    export PATH="${HOME}/.local/bin:${PATH}"
+    echo "gitleaks installed"
+    return 0
+  fi
+  echo "::warning::Failed to install gitleaks — secret scans will be skipped"
+  return 1
+}
```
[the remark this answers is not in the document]
subhashkhileri: Fixed — `add_label` now captures stderr and emits `::warning::` on failure.

.fullsend/rhdh/harness/e2e-triage.yaml
```diff
@@ -0,0 +1,67 @@
...
+post_script: rhdh/scripts/post-e2e-triage.sh
+
+validation_loop:
+  script: rhdh/scripts/validate-output-schema.sh
+  max_iterations: 2
+
+env:
+  runner:
+    GH_TOKEN: ${GH_TOKEN}
+    REPO_FULL_NAME: redhat-developer/rhdh-plugin-export-overlays
```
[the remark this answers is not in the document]
subhashkhileri: Fixed — added `GITHUB_ISSUE_URL` to `env.runner`.

.fullsend/rhdh/agents/e2e-triage.md
```diff
@@ -0,0 +1,500 @@
...
+## Input
+
+This agent is triggered by a GitHub issue labeled `e2e-triage`. The issue
+body contains the prow URL. Extract it on startup:
+
+```bash
+ISSUE_URL="${GITHUB_ISSUE_URL:-}"
+if [[ -z "${ISSUE_URL}" ]]; then
+  echo "ERROR: GITHUB_ISSUE_URL is not set" >&2
+  exit 1
```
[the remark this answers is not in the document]
subhashkhileri: Acknowledged — mitigated by sandbox disallowedTools and read-only network policy; narrow `grep -oP` extraction limits surface.

.fullsend/rhdh/scripts/post-e2e-triage.sh
```diff
@@ -0,0 +1,276 @@
...
+
+# add_label uses the labels API to avoid firing issues.edited.
+add_label() {
+  local repo="$1" issue="$2" label="$3"
```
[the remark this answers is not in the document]
subhashkhileri: Harness `validation_loop` validates against the JSON schema before the post-script runs — duplicating in bash is unnecessary.

.fullsend/rhdh/scripts/post-e2e-triage.sh
```diff
@@ -0,0 +1,276 @@
...
+for dir in iteration-*/output; do
+  if [[ -f "${dir}/agent-result.json" ]]; then
+    RESULT_FILE="${dir}/agent-result.json"
+  fi
+done
+
+if [[ -z "${RESULT_FILE}" ]]; then
+  echo "::warning::No agent-result.json found"
+  ls -R iteration-*/ 2>/dev/null || true
+  exit 0
```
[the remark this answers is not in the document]
subhashkhileri: Coder has `cancel-in-progress` concurrency groups — duplicate runs self-resolve.

.fullsend/rhdh/schemas/e2e-triage-result.schema.json
```diff
@@ -0,0 +1,135 @@
...
+      "additionalProperties": false,
+      "required": ["action"],
+      "properties": {
+        "action": {
+          "type": "string",
+          "enum": ["create", "comment", "skip"],
+          "description": "Whether to create a new issue, comment on an existing one, or skip."
+        },
+        "number": {
+          "type": ["integer", "null"],
```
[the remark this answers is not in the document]
subhashkhileri: Functionally harmless — `action` is already required at the top level.

.fullsend/rhdh/scripts/post-e2e-triage.sh
```diff
@@ -0,0 +1,276 @@
...
+        remove_label "${REPO_FULL_NAME}" "${ISSUE_NUMBER}" "ready-to-code"
+        sleep 1
+        add_label "${REPO_FULL_NAME}" "${ISSUE_NUMBER}" "ready-to-code"
+        echo "  ready-to-code label cycled — coder will re-trigger"
+      fi
+      ;;
+
+    skip)
+      echo "  No issue directive — skipping"
+      ;;
```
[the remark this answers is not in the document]
subhashkhileri: Values come from agent JSON output in a sandboxed environment — low risk.

.fullsend/rhdh/scripts/post-e2e-triage.sh
```diff
@@ -0,0 +1,276 @@
...
+export GH_TOKEN
+echo "::add-mask::${GH_TOKEN}"
+
+PUSH_TOKEN="${PUSH_TOKEN:-${GH_TOKEN}}"
+echo "::add-mask::${PUSH_TOKEN}"
+
+# Promote to PUSH_TOKEN for write permissions on issues and labels.
+export GH_TOKEN="${PUSH_TOKEN}"
+
+# Extract trigger issue number from GITHUB_ISSUE_URL
```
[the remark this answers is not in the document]
subhashkhileri: `gh` CLI rejects non-numeric input — explicit validation unnecessary.

.fullsend/rhdh/scripts/post-e2e-triage.sh
```diff
@@ -0,0 +1,292 @@
...
+  fi
+  echo "::warning::Failed to install gitleaks — secret scans will be skipped"
+  return 1
+}
+
+# add_label uses the labels API to avoid firing issues.edited.
+add_label() {
+  local repo="$1" issue="$2" label="$3"
+  gh api "repos/${repo}/issues/${issue}/labels" -f "labels[]=${label}" --silent 2>/dev/null || true
+}
```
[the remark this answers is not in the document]
subhashkhileri: Fixed in the medium-severity duplicate — `add_label` now emits `::warning::` on failure.

.fullsend/rhdh/agents/e2e-triage.md
```diff
@@ -0,0 +1,490 @@
...
+name: e2e-triage
+description: >-
+  Analyze E2E nightly test failures, classify root causes per workspace,
+  search for existing issues (dedup), and emit structured issue directives.
+  Does NOT modify code, create branches, or fix tests.
+model: opus
+disallowedTools: >-
+  Bash(git push *), Bash(git push),
+  Bash(git checkout -b *), Bash(git checkout -b),
+  Bash(git add *), Bash(git add),
```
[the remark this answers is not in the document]
subhashkhileri: See reply on the medium-severity duplicate — agent needs `gh api` for read-only dedup/search.

.fullsend/rhdh/schemas/e2e-triage-result.schema.json
```diff
@@ -0,0 +1,135 @@
...
+          "description": "Issue title (required for create action)."
+        },
+        "labels": {
+          "type": "array",
+          "items": { "type": "string" },
+          "maxItems": 10,
+          "description": "Labels for new issue."
+        },
+        "body": {
+          "type": "string",
```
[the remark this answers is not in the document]
subhashkhileri: See reply on the duplicate — `action` is required at the top level, making conditional `required` redundant.

.fullsend/rhdh/agents/e2e-triage.md
```diff
@@ -0,0 +1,523 @@
...
+7. Prow URL:
+   ```
+   ## Artifacts
+
+   <prow URL>
+   ```
+
+### Umbrella issue (≥3 workspaces with same root cause)
+
+When ≥3 workspaces share the same `root_cause_slug`, create ONE umbrella
```
gashcrumb: The prose says create ONE umbrella issue for ≥3 workspaces, but the schema and post-script are strictly per-workspace. There's no emit rule like "`create` on the first workspace, `skip` (or `comment`) on the siblings." Easy failure mode if the model emits `create` three times: duplicate umbrella issues and concurrent coder runs on overlapping workspaces.

Worth spelling out the sibling actions explicitly in Phase 5/6 (and maybe noting it in the schema description for `issue.action`).

.fullsend/rhdh/scripts/post-e2e-triage.sh
```diff
@@ -0,0 +1,286 @@
...
+  exit 1
+fi
+
+echo "Target branch: $(jq -r '.target_branch // "main"' "${RESULT_FILE}")"
+echo "Workspaces to process: ${WORKSPACE_COUNT}"
+
+# ---------------------------------------------------------------------------
+# 2. Scan agent-result.json for secrets
+# ---------------------------------------------------------------------------
+if install_gitleaks; then
```
gashcrumb: If `install_gitleaks` fails, this block is skipped and the script continues posting agent-written issue bodies/comments to a public repo. That's the only secret gate before publish.

Prefer fail-closed here — e.g. if install fails, `exit 1` (or skip all create/comment steps) rather than warn-and-continue.

.fullsend/rhdh/agents/e2e-triage.md
```diff
@@ -0,0 +1,523 @@
...
+| `environment` | `e2e-failure` | No |
+| `infra_flake` | (no issue) | — |
+
+**Issue title format:** `[fullsend] E2E: <workspace> — <short description>`
+
+**Issue body must include:**
+
+1. **Tracking line** (first line, visible, searchable):
+   ```
+   `fullsend-tracking: workspace=<name> root-cause=<slug> branch=<branch>`
```
gashcrumb: Release-branch handoff looks fragile compared to the old `e2e-fix` post-script, which passed `target_branch` straight into `gh pr create --base`. Here the branch only lives in a tracking line, and the scaffold coder has to notice it from issue context.

The upstream code-implementation skill *can* pick a branch from the issue, but a missed read means release-nightly fixes land on `main`. Consider making the remediation section state the target branch explicitly ("Open the PR against `release-x.y`"), or verifying once with a release-branch dry run.

.fullsend/rhdh/agents/e2e-triage.md
```diff
@@ -0,0 +1,523 @@
...
+
+# 2. If found, check if it has a linked PR
+WITH_PR=$(curl -sf "https://api.github.com/search/issues" \
+  --data-urlencode "q=repo:${REPO} is:issue state:open linked:pr \"fullsend-tracking: workspace=${WORKSPACE}\" in:body" \
+  | jq '[.items[] | {number, title}]')
+```
+
+### Decision matrix
+
+| Issue found | Has linked PR | Action |
```
gashcrumb: `linked:pr` matches issues linked to open, closed, or merged PRs. The matrix treats "has linked PR" as "coder already working," but `cycle_ready_to_code` was designed for the closed-without-merge case — which still looks like "has linked PR" to search.

Suggest checking linked PR state (open vs closed/merged) before deciding whether to cycle, e.g.:

| Issue found | Linked PR open | Action |
|---|---|---|
| Yes | Yes | comment only |
| Yes | No / closed / merged | comment + `cycle_ready_to_code` |

AGENTS.md
```diff
@@ -443,6 +443,31 @@ Two Claude Code skills are available at `.claude/skills/` for investigating E2E
...
+### Prohibited modifications
+- Plugin source code (`workspaces/*/plugins/`)
+- CI configuration (`.github/`)
+- Repository config (`CLAUDE.md`, `CODEOWNERS`, `.fullsend/`)
+
+### Skipping tests (product_bug classification)
+When the issue says `fix_category: product_bug`, add `test.skip` instead
+of fixing the test:
+
+    test.skip(!!process.env.E2E_NIGHTLY_MODE, "<root cause summary>");
```
gashcrumb, line 462: This simplified pattern drops the `JOB_NAME?.includes("periodic-")` half that the old agent (and `workspaces/backstage/.../auth.spec.ts`) used. Prow currently exports `E2E_NIGHTLY_MODE=true` for the overlays nightly, so it likely works today — but `!!process.env.E2E_NIGHTLY_MODE` is also truthy for the string `"false"`, unlike the `=== "true" || === "1"` checks in lightspeed/intelligent-assistant.

Prefer aligning with the stronger existing pattern so skip PRs don't depend on one CI env var staying set forever.

.fullsend/rhdh/scripts/post-e2e-triage.sh
```diff
@@ -0,0 +1,286 @@
...
+        echo "  Created issue #${ISSUE_NUMBER}: ${ISSUE_URL}"
+        ISSUE_REF="#${ISSUE_NUMBER}"
+
+        # Add labels via labels API. Batch non-deferred labels into one
+        # call, then add ready-to-code LAST so its labeled event is the
+        # final webhook and triggers the coder.
+        NON_DEFERRED="$(jq -c "[.workspaces[$i].issue.labels // [] | .[] | select(. != \"ready-to-code\")]" "${RESULT_FILE}")"
+        HAS_DEFERRED="$(jq -r ".workspaces[$i].issue.labels // [] | map(select(. == \"ready-to-code\")) | length > 0" "${RESULT_FILE}")"
+        if [[ "${NON_DEFERRED}" != "[]" ]]; then
+          echo "${NON_DEFERRED}" | gh api "repos/${REPO_FULL_NAME}/issues/${ISSUE_NUMBER}/labels" --input - --silent 2>/dev/null || true
```
gashcrumb: Nit: `add_label` now emits `::warning::` on failure, but this batch path for non-deferred labels is still `2>/dev/null || true` with no log. Same silent-failure class if `e2e-failure` (etc.) can't be applied.

gashcrumb (commented):
Architecture looks sound — triage → scaffold coder is a cleaner split than the monolith. A few things I'd want tightened or verified before relying on this in production nightlies.

Main asks: explicit umbrella emit rules in the per-workspace schema, fail-closed gitleaks before posting issue bodies, and confirmation that release-branch targeting survives the handoff to the generic coder. The test plan is still unchecked end-to-end, which is the biggest remaining risk.

.fullsend/rhdh/agents/e2e-triage.md
subhashkhileri: Good catch — updated Phase 5 to spell out the umbrella emit rules explicitly: when ≥3 workspaces share a `root_cause_slug`, emit ONE entry (not per-workspace) with `workspace` = the slug, combined `tests`, and per-workspace sections in the body. Sibling workspaces are NOT emitted as separate entries. Also updated Phase 6 field rules to document this.

.fullsend/rhdh/scripts/post-e2e-triage.sh
subhashkhileri: Fixed — `install_gitleaks` failure now exits the script instead of skipping the scan.

.fullsend/rhdh/agents/e2e-triage.md
subhashkhileri: Added explicit `**Target branch:** \`<TARGET_BRANCH>\`` to the remediation section template. Also confirmed working end-to-end: issue #2984 → scaffold coder → PR #2990 targeting `release-1.10`.

.fullsend/rhdh/agents/e2e-triage.md
subhashkhileri: Good catch — replaced `linked:pr` search with timeline API to check actual PR state. Decision matrix now distinguishes open vs closed/merged linked PRs.

AGENTS.md
subhashkhileri: Kept `!!process.env.E2E_NIGHTLY_MODE` — CI sets it to `"true"` or leaves it unset, never `"false"`. Simpler and matches current CI behavior.

.fullsend/rhdh/scripts/post-e2e-triage.sh
subhashkhileri: Fixed — batch label add now emits `::warning::` on failure, matching the `add_label` helper.

.fullsend/rhdh/schemas/e2e-triage-result.schema.json
```diff
@@ -0,0 +1,135 @@
...
+          "type": ["integer", "null"],
+          "description": "Existing issue number (required for comment action)."
+        },
+        "title": {
+          "type": "string",
+          "minLength": 1,
+          "maxLength": 256,
+          "description": "Issue title (required for create action)."
+        },
+        "labels": {
```
gustavolira, line 106: There's no allowlist on this. If the agent ever puts `e2e-triage` in here — the label that triggered it, which appears all over its own prompt — the post-script applies it to the newly created issue, the CEL trigger in `e2e-triage.yaml` matches, and triage runs again and files more issues. Since `fix_category` and `action` are already enums, could we do the same here?

```json
"labels": {
  "type": "array",
  "items": { "enum": ["e2e-failure", "ready-to-code"] },
  "maxItems": 10
}
```

.fullsend/rhdh/scripts/post-e2e-triage.sh
```diff
@@ -0,0 +1,290 @@
...
+  )
+
+  echo ""
+  echo "--- Workspace: ${WS_NAME} (${FIX_CAT}) ---"
+
+  ISSUE_REF=""
+
+  case "${ISSUE_ACTION}" in
+    create)
+      echo "  Creating GitHub issue..."
```
gustavolira, line 195: Dedup lives entirely in the agent's Phase 4 search, and this path creates unconditionally. When that search misses — exact-phrase matching on `"fullsend-tracking: workspace=X" in:body` isn't especially reliable, and the agent is non-deterministic on top of it — you get a fresh issue plus `ready-to-code`, so another coder run and another PR, every night for the same failure. The old flow at least capped at 2 attempts and bailed when a fix PR was already open.

Could we add a deterministic guard right here: search open issues for the same title and downgrade to a comment on a hit? And maybe a ceiling on how many creates execute per invocation — the schema allows 30 workspaces, which is 30 coder runs out of one bad nightly.

.fullsend/rhdh/schemas/e2e-triage-result.schema.json
```diff
@@ -0,0 +1,135 @@
...
+        }
+      },
+      "allOf": [
+        {
+          "if": { "properties": { "action": { "const": "create" } } },
+          "then": { "required": ["title", "body"] }
+        },
+        {
+          "if": { "properties": { "action": { "const": "comment" } } },
+          "then": { "required": ["number", "body"] }
```
gustavolira, line 130: `number` is typed `["integer", "null"]`, so `required: ["number"]` here is satisfied by an explicit `null`. The agent's own output template in `e2e-triage.md` shows `"number": null` sitting right next to `"action": "<create|comment|skip>"`, so it's a pretty likely emission — and it lands on `gh issue comment null`, which only warns and drops the analysis on the floor. Either drop `null` from the type and make the field optional, or add `"not": { "properties": { "number": { "const": null } } }` to this branch.

.fullsend/rhdh/scripts/post-e2e-triage.sh
```diff
@@ -0,0 +1,290 @@
...
+
+      # ---------------------------------------------------------------
+      # 4. Handle cycle_ready_to_code
+      # ---------------------------------------------------------------
+      if [[ "${CYCLE}" == "true" ]]; then
+        echo "  Cycling ready-to-code label on #${ISSUE_NUMBER}..."
+        remove_label "${REPO_FULL_NAME}" "${ISSUE_NUMBER}" "ready-to-code"
+        sleep 1
+        add_label "${REPO_FULL_NAME}" "${ISSUE_NUMBER}" "ready-to-code"
+        echo "  ready-to-code label cycled — coder will re-trigger"
```
gustavolira: `remove_label` swallows everything with `2>/dev/null || true`, so if the removal fails the label is still on the issue, re-adding an existing label doesn't fire a fresh `labeled` event, and the coder never re-triggers — but this line says it will. Can we capture the removal result and only claim the cycle when it actually happened?

.fullsend/rhdh/scripts/post-e2e-triage.sh
```diff
@@ -0,0 +1,290 @@
...
+#
+# Optional environment variables:
+#   PUSH_TOKEN        — dedicated token with issues:write (falls back to GH_TOKEN)
+set -euo pipefail
+
+# ---------------------------------------------------------------------------
+# Configuration
+# ---------------------------------------------------------------------------
+GITLEAKS_VERSION="8.30.1"
+GITLEAKS_SHA256="551f6fc83ea457d62a0d98237cbad105af8d557003051f41f3e7ca7b3f2470eb"
```
gustavolira, line 34: This checksum is specifically the one for `gitleaks_8.30.1_linux_x64.tar.gz` (checked it against the release checksums file), but lines 78-84 detect os/arch and build a per-platform URL — so the `aarch64|arm64` branch can never pass `sha256sum -c`. That was survivable when a failed install was just a warning; now that it's `exit 1` at line 158, the whole post-script dies on any non-x64 runner and no issues get filed at all. Either pin `linux_x64` explicitly and drop the arch switch, or map the checksums:

```bash
declare -A GITLEAKS_SHA256=(
  [linux_x64]="551f6fc8…"
  [linux_arm64]="e4a487ee…"
)
```

The hard-fail itself is the right call — it just needs the arch handling to match.

.fullsend/rhdh/scripts/post-e2e-triage.sh
```diff
@@ -0,0 +1,290 @@
...
+      fi
+
+      # Create issue WITHOUT labels — labels are added separately via the
+      # labels API so that ready-to-code fires a proper issues.labeled event
+      # (gh issue create --label does NOT emit a separate labeled event).
+      if ISSUE_URL="$(gh issue create \
+        --repo "${REPO_FULL_NAME}" \
+        --title "${ISSUE_TITLE}" \
+        --body "${ISSUE_BODY}" 2>&1)"; then
+        ISSUE_NUMBER="$(echo "${ISSUE_URL}" | grep -o '[0-9]*$')"
```
gustavolira: `2>&1` folds stderr into the captured value and `grep -o '…$'` matches per line, so a single extra line ending in a digit gives you a multi-line `ISSUE_NUMBER` that goes straight into the API path at line 212. This file already does it the clean way at line 52 (`${TRIGGER_ISSUE_URL##*/}`), same as the workflow at line 147 — could we redirect stderr to a temp file the way `add_label` does, use `${ISSUE_URL##*/}`, and reject anything that isn't `^[0-9]+$` before building the URL? Worth splitting the error text into its own variable too, since `ISSUE_URL` currently holds gh's stderr on the failure path.

.fullsend/rhdh/scripts/post-e2e-triage.sh
```diff
@@ -0,0 +1,290 @@
...
+rm -rf "${SCAN_DIR}"
+echo "Result file scan passed"
+
+# ---------------------------------------------------------------------------
+# 3. Execute issue directives per workspace
+# ---------------------------------------------------------------------------
+declare -a SUMMARY_LINES=()
+
+for i in $(seq 0 $((WORKSPACE_COUNT - 1))); do
+  read -r WS_NAME FIX_CAT TEST_COUNT ISSUE_ACTION < <(
```
gustavolira: `read` without `IFS=$'\t'` splits on spaces as well as tabs, and `workspace` in the schema is any string up to 128 chars — which now includes umbrella entries where the agent invents the value. One space shifts every field one to the left and the `case` below falls through to the unknown-action warning. `IFS=$'\t' read -r …` covers it.

.fullsend/rhdh/scripts/post-e2e-triage.sh
```diff
@@ -0,0 +1,290 @@
...
+  for line in "${SUMMARY_LINES[@]}"; do
+    SUMMARY+="${line}\n"
+  done
+
+  AGENT_SUMMARY="$(jq -r '.summary // empty' "${RESULT_FILE}")"
+  if [[ -n "${AGENT_SUMMARY}" ]]; then
+    SUMMARY+="\n### Analysis\n\n${AGENT_SUMMARY}"
+  fi
+
+  printf '%b' "${SUMMARY}" | gh issue comment "${TRIGGER_ISSUE_NUMBER}" \
```
gustavolira: `%b` also interprets escapes inside `AGENT_SUMMARY`, which is agent-authored prose — a root cause mentioning a Windows path or a regex like `\d` comes out mangled. Building the body with real newlines and piping it in sidesteps that:

```bash
{
  echo "## Triage Summary"; echo
  echo "| Workspace | Category | Tests | Issue |"
  echo "|-----------|----------|-------|-------|"
  printf '%s\n' "${SUMMARY_LINES[@]}"
} | gh issue comment "${TRIGGER_ISSUE_NUMBER}" --repo "${REPO_FULL_NAME}" --body-file -
```

.fullsend/rhdh/harness/e2e-triage.yaml
```diff
@@ -0,0 +1,68 @@
...
+    GITHUB_ISSUE_URL: "${GITHUB_ISSUE_URL}"
+    REPO_FULL_NAME: redhat-developer/rhdh-plugin-export-overlays
+
+timeout_minutes: 60
+
+forge:
+  github:
+    env:
+      runner:
+        PUSH_TOKEN: "${PUSH_TOKEN}"
```
gustavolira, line 65: Following up on the `GITHUB_ISSUE_URL` fix above — the top-level `env.runner` has it now, but this `forge.github.env.runner` block still doesn't, while `forge.github.env.sandbox` re-declares all three of its vars. If forge blocks merge with the top level, that sandbox repetition is dead weight; if they replace it, the runner loses `GITHUB_ISSUE_URL` and the summary-comment step silently no-ops on the guard at line 265. Which way does it work?

.fullsend/rhdh/policies/e2e-triage.yaml
```diff
@@ -1,12 +1,13 @@
...
-# Sandbox policy for the e2e-fix agent.
+# Sandbox policy for the e2e-triage agent.
 #
-# Based on rhdh/policies/code.yaml (PR #2794) with additions:
-#   - github_api is read-only (writes happen in the post-script on the runner)
-#   - prow_artifacts policy for downloading nightly CI artifacts
-#   - **/gh binary for GitHub read operations (issue view, pr list, dedup)
+# Based on rhdh/policies/e2e-fix.yaml with the following differences:
+#   - package_registries removed (triage doesn't build or install packages)
+#   - github_api remains read-only (writes happen in the post-script)
```
gustavolira, line 9: Dropping the registries makes sense for an agent that doesn't build, but `playwright-trace/SKILL.md` declares `allowed-tools: Bash(npx:*)` and documents falling back on `npx` when the global binary isn't available — that path now hangs instead of degrading. Since `e2e-triage.md` makes trace inspection a hard requirement (subagents must exit if the skill can't invoke), worth confirming `/usr/bin/playwright` is actually present in `rhdh-fullsend-code:latest` so the fallback is never needed.

.fullsend/rhdh/agents/e2e-triage.md
```diff
@@ -0,0 +1,570 @@
...
+   ```
+
+### Umbrella issue (≥3 workspaces with same root cause)
+
+When ≥3 workspaces share the same `root_cause_slug`, emit **ONE entry**
+in the `workspaces` array — not one per workspace. **Do NOT emit separate
+entries for sibling workspaces covered by the umbrella.**
+
+**Structured output rules for umbrella entries:**
+- `workspace`: use the `root_cause_slug` (e.g., `oci-resolution`), not a
```
gustavolira: nit: this overloads `workspace` — the schema documents it as "Workspace name (e.g., argocd, orchestrator, backstage)", and the post-script prints it under a "Workspace" column in the summary table with nothing marking it as an umbrella. An explicit `"umbrella": true` (or `"covers": ["argocd", "tekton", …]`) would keep the field honest and let the summary render something like "umbrella (3 workspaces)".

.fullsend/rhdh/schemas/e2e-triage-result.schema.json
```diff
@@ -0,0 +1,135 @@
...
+    "target_branch": {
+      "type": "string",
+      "minLength": 1,
+      "maxLength": 128,
+      "description": "Branch the nightly ran against (e.g., main, release-1.10)."
+    },
+    "workspaces": {
+      "type": "array",
+      "minItems": 1,
+      "maxItems": 30,
```
gustavolira: nit: these caps bite hardest in exactly the run you most want triaged — the nightly where a bad chart bump fails 60 tests in one workspace is the one the schema rejects, and after `max_iterations: 2` you get nothing at all. Either raise them substantially or have the agent truncate `tests` with an explicit "and N more" note so validation never blocks the run.

.fullsend/rhdh/agents/e2e-triage.md
```diff
@@ -0,0 +1,570 @@
...
+   <specific files to modify, what to change, what pattern to follow>
+   ```
+   Always state the target branch explicitly so the scaffold coder opens
+   the PR against the correct branch. The scaffold coder reads this issue
+   and implements the fix. Vague instructions like "fix the timeout"
+   produce vague fixes. Instead:
+   "In `workspaces/argocd/e2e-tests/tests/specs/argocd.spec.ts` line 42,
+   increase the route wait timeout from 30s to 60s."
+
+6. For `product_bug` — explicit skip instruction:
```
gustavolira: Dropping JIRA from the pipeline means a product bug now gets a `test.skip` plus an issue in the overlays repo, and nothing in the tracker the plugin team actually watches — the old flow filed an RHDHBUGS bug and backfilled the key into the skip reason. The skipped test then has no owner and nothing that triggers a re-enable. Is the plan to file those by hand, or should this remediation block carry a "file an RHDHBUGS bug and put the key in the skip reason" step?

AGENTS.md
```diff
@@ -443,6 +443,31 @@ Two Claude Code skills are available at `.claude/skills/` for investigating E2E
...
+
+When fixing E2E test failures from `[fullsend] E2E:` issues:
+
+### Allowed modifications
+- `workspaces/<workspace>/e2e-tests/` — any file under the e2e-tests directory
+
+### Prohibited modifications
+- Plugin source code (`workspaces/*/plugins/`)
+- CI configuration (`.github/`)
+- Repository config (`CLAUDE.md`, `CODEOWNERS`, `.fullsend/`)
```
gustavolira, line 456: nit: `CLAUDE.md` in this repo is a one-line `@AGENTS.md` include, so all the actual content lives in the file that isn't on this list. Worth adding `AGENTS.md` here.

.fullsend/rhdh/scripts/post-e2e-triage.sh
```diff
@@ -0,0 +1,290 @@
...
+    skip)
+      echo "  No issue directive — skipping"
+      ;;
+
+    *)
+      echo "::warning::Unknown issue action '${ISSUE_ACTION}' for ${WS_NAME} — skipping"
+      ;;
+  esac
+
+  SUMMARY_LINES+=("| ${WS_NAME} | \`${FIX_CAT}\` | ${TEST_COUNT} | ${ISSUE_REF:-—} |")
```
gustavolira: nit: for `infra_flake` this summary comment is the only record that ever exists, and the row reads `| orchestrator | infra_flake | 3 | — |`. Adding `root_cause` (or `root_cause_slug`, which the schema requires but nothing in this script uses) would make those rows worth reading, and would let someone scanning a week of trigger issues spot a "flake" that recurs every single night.

.fullsend/rhdh/agents/e2e-triage.md
gashcrumb: Thanks — that dry-run is convincing. Confirmed #2984 → #2990 with base `release-1.10`, so the tracking-line + `**Target branch:**` handoff works for the normal `test_fix` path.

One leftover: the `product_bug` remediation template (item 6) still replaces the whole Remediation block and drops `**Target branch:**`. Same for the per-workspace `### Remediation` sections in the umbrella body template. Skip PRs against release branches would regress to the soft handoff again unless those templates get the same line.

AGENTS.md
gashcrumb: Fair enough — if CI only ever sets `"true"` or leaves it unset, `!!process.env.E2E_NIGHTLY_MODE` is fine. I'm good leaving this as-is given that constraint (and the verified dry-run). Worth a one-line note in AGENTS.md if anyone later defaults it to `"false"` in `run-e2e.sh` exports, but not blocking.

.fullsend/rhdh/agents/e2e-triage.md
gashcrumb: Looks good — thanks for spelling that out.

.fullsend/rhdh/scripts/post-e2e-triage.sh
gashcrumb: Perfect, thanks.

.fullsend/rhdh/agents/e2e-triage.md
gashcrumb: Nice — the open-vs-closed distinction is what I was after.

.fullsend/rhdh/scripts/post-e2e-triage.sh
gashcrumb: Thanks.

.fullsend/rhdh/agents/e2e-triage.md
```diff
@@ -0,0 +1,570 @@
...
+   ```
+   ## Remediation
+
+   This is a product bug — do not fix the test. Instead, add `test.skip`:
```
gashcrumb: The general remediation template above now includes `**Target branch:**`, but this `product_bug` block replaces the whole Remediation section and omits it. Same gap in the umbrella body's per-workspace `### Remediation` sections.

Since skip PRs still go through the scaffold coder (with `ready-to-code`), release-branch product bugs would lose the explicit base-branch signal that #2990 just proved matters. Worth adding `**Target branch:** \`<TARGET_BRANCH>\`` here too.

gashcrumb (commented):
Follow-up after the latest fixes and author replies.

Most of my earlier comments look addressed — including a real handoff dry-run (#2984 → #2990 on `release-1.10`). Leaving one small template gap on `product_bug` / umbrella remediation (inline). Gustavo's review already covers the stronger remaining correctness items (label allowlist, create-path dedup guard, `number: null`, etc.) better than I'd restate them.

.fullsend/rhdh/agents/e2e-triage.md
```diff
@@ -0,0 +1,570 @@
...
+
+### Umbrella issue search
+
+When ≥3 workspaces share the same `root_cause_slug`, search for an existing
+umbrella issue:
+
+```bash
+ROOT_CAUSE_SLUG="<slug>"
+gh api -X GET search/issues \
+  -f q="repo:${REPO} is:issue state:open \"fullsend-tracking: root-cause=${ROOT_CAUSE_SLUG}\" in:body" \
```
gustavolira, line 306: This search can never match anything. The tracking line the agent is told to emit (line 332, and the umbrella example at 409-411) is:

```
fullsend-tracking: workspace=argocd root-cause=oci-resolution branch=main
```

so the literal phrase `fullsend-tracking: root-cause=<slug>` never appears in any issue body — `root-cause=` is always preceded by `workspace=<name> `. The per-workspace search at line 274 works precisely because `workspace=` is the first field. Net effect: umbrella issues never dedup, so a fresh umbrella issue plus a fresh coder run every night for the same root cause.

The underlying problem is that the tracking-line grammar is authored twice, 40 lines apart, with no single definition — and the two copies are already inconsistent in v1. Worth defining it once before Phase 4 and making each key independently searchable:

```
`fullsend-tracking: workspace=<name>`
`fullsend-tracking: root-cause=<slug>`
`fullsend-tracking: branch=<branch>`
```

Then both queries are prefix matches on the same documented token, and adding a fourth key later can't silently break either one.

.fullsend/rhdh/scripts/post-e2e-triage.sh
```diff
@@ -0,0 +1,290 @@
...
+TRIGGER_ISSUE_NUMBER=""
+if [[ -n "${TRIGGER_ISSUE_URL}" ]]; then
+  TRIGGER_ISSUE_NUMBER="${TRIGGER_ISSUE_URL##*/}"
+fi
+
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+
+sanitize_for_gha() {
```
gustavolira, line 59: This is single-pass, so it can be walked straight through. I ran it:

```
in:  %%0A0A    out: %0A    <== reconstituted
in:  %0%0AA    out: %0A    <== reconstituted
in:  %%0a0a    out: %0a    <== reconstituted
```

`${text//%0A/}` removes non-overlapping matches in one sweep, so a crafted string re-forms the exact sequence the function exists to strip. It's applied to `gh` stderr (lines 105 and 220), which echoes back agent-authored titles and label names — so agent output can land a `%0A` in a `::warning::` line and forge additional workflow commands.

Looping until stable fixes it:

```bash
while [[ "${text}" != "${prev}" ]]; do prev="${text}"; text="${text//%0A/}"; …; done
```

Though honestly the cleaner move is to stop hand-rolling this and write untrusted text to `$GITHUB_STEP_SUMMARY` instead of into a workflow command.

.fullsend/rhdh/agents/e2e-triage.md
```diff
@@ -0,0 +1,570 @@
+---
+name: e2e-triage
+description: >-
+  Analyze E2E nightly test failures, classify root causes per workspace,
+  search for existing issues (dedup), and emit structured issue directives.
+  Does NOT modify code, create branches, or fix tests.
+model: opus
+disallowedTools: >-
```
gustavolira, line 8: The harness says the triage agent "does NOT modify code" (`e2e-triage.yaml:8-9`) and the Constraints section repeats it at line 536, but nothing enforces it — this list only covers `Bash(git …)` and `Bash(gh …)`. `Edit`, `Write`, and `MultiEdit` are unrestricted, and the policy keeps `include_workdir: true` with `/sandbox` read-write, so the checkout is writable.

The old harness was explicit that this list is the mechanism, not the prose: *"The agent NEVER pushes or creates PRs (disallowedTools enforces this)"*. For the fix agent writes were the point; here a stray edit is worse than harmless — nothing commits or pushes, so it vanishes with the sandbox and silently corrupts the analysis that everything downstream depends on. Adding `Edit, Write, MultiEdit` here (or mounting the workdir read-only) makes the claim true.

.fullsend/rhdh/policies/e2e-triage.yaml
```diff
@@ -1,12 +1,13 @@
...
+#   - prow_artifacts retained for downloading nightly CI artifacts
+#   - **/gh binary for GitHub read operations (issue search, dedup)
```
gustavolira, line 11: This line says `**/gh` for GitHub read operations, but `github_api.binaries` at line 54 still grants `**/curl` as well — and the sibling policy documents that exact grant as deliberately forbidden. `policies/code.yaml:16-18`:

```
#   - GitHub API (gh/git only — curl intentionally excluded to limit raw
#     HTTP access with the injected GH_TOKEN)
```

`code.yaml:65-69` lists only `gh`, `git`, `node`. Meanwhile `harness/e2e-triage.yaml:53` injects `GH_TOKEN` into the sandbox — the combination that comment exists to prevent.

It's dead weight too: every GitHub call in the agent prompt uses `gh` (`:36`, `:273`, `:282`, `:305`), not curl. The only thing still pointing at curl is the stale prose at `e2e-triage.md:114` ("Read GitHub issues, PRs, labels via `curl` + GitHub REST API"), inherited from `e2e-fix.md` where it was accurate — and this PR dropped that file's "public repo, no auth needed" qualifier while keeping the grant. Could we drop `**/curl` from `github_api` (keep it under `prow_artifacts`, which genuinely needs it) and fix line 114 to say `gh`?

.fullsend/rhdh/harness/e2e-triage.yaml
```diff
@@ -0,0 +1,68 @@
...
+
+role: triage
+slug: fullsend-ai-triage
+
+trigger: |
+  event.entity.kind == "work_item"
+  && event.transition.kind == "label_changed"
+  && event.transition.label.name == "e2e-triage"
+
+host_files:
```
gustavolira, line 24: This drops the `rhdh/env/rhdh-toolchain.env` mount that `code.yaml:5-9`, `fix.yaml:5-9`, and the deleted `e2e-fix.yaml:32-35` all carry. Dropping `yarn-proxy.env` tracks with removing `package_registries` — triage installs nothing — but the toolchain file also carries this, and the reason is written down in it:

```
# openspec — disable telemetry to avoid DENIED requests to edge.openspec.dev
# (the sandbox policy does not allowlist this endpoint).
```

The image is unchanged (line 13, same `rhdh-fullsend-code:latest`), so `OPENSPEC_TELEMETRY=0` is now unset and `edge.openspec.dev` isn't in this agent's policy — which reintroduces exactly the DENIED-request noise that env file exists to suppress. That's image-level, not build-related, so it should survive the registry cleanup. Re-add the mount, or split the openspec stanza into a triage-specific env file.

.fullsend/rhdh/policies/e2e-triage.yaml
```diff
@@ -1,12 +1,13 @@
...
 version: 1
 
-# Sandbox policy for the e2e-fix agent.
+# Sandbox policy for the e2e-triage agent.
 #
-# Based on rhdh/policies/code.yaml (PR #2794) with additions:
-#   - github_api is read-only (writes happen in the post-script on the runner)
-#   - prow_artifacts policy for downloading nightly CI artifacts
-#   - **/gh binary for GitHub read operations (issue view, pr list, dedup)
+# Based on rhdh/policies/e2e-fix.yaml with the following differences:
```
gustavolira: This points at `rhdh/policies/e2e-fix.yaml`, which this same PR deletes — it *is* this file, renamed. The provenance pointer is dangling the moment it merges. `policies/code.yaml:6-8` establishes the format for these headers:

```
# forked-from: fullsend v0.17.0 scaffold
# last-synced: 2026-06-16
# Based on upstream code policy with additions:
```

Could we point at a file that survives (`code.yaml`) and add the `forked-from:` / `last-synced:` lines so this fork can actually be re-synced later?

.github/workflows/e2e-triage-agent.yaml
```diff
@@ -124,8 +129,8 @@ jobs:
           fi
 
           # Create the label if it doesn't exist
-          gh label create "e2e-fix-agent" --repo "${REPO}" \
-            --description "Triggers the fullsend e2e-fix agent" --color "7057ff" \
+          gh label create "e2e-triage" --repo "${REPO}" \
```
gustavolira, line 132: This workflow mints the `e2e-triage` label that starts a privileged agent run, but it isn't covered by the CODEOWNERS protection `docs/fullsend.md:174-176` advertises for the fullsend surface — `CODEOWNERS:85-87` covers `.fullsend/` and `.github/workflows/fullsend.yaml`, and this file only matches the catch-all `*` at line 17. So the trigger for the agent is less protected than the agent itself.

Pre-existing (`e2e-fix-agent.yaml` had the same gap), but renaming the file is the natural moment to close it — one line in CODEOWNERS.

.fullsend/rhdh/agents/e2e-triage.md
```diff
@@ -0,0 +1,570 @@
...
+---
+
+## Phase 5: Emit Directives
+
+For each workspace, write an issue directive based on the classification
+and dedup results.
+
+### Per-workspace issue (≤2 workspaces with same root cause)
+
+| Category | Labels | `ready-to-code` |
```
gustavolira: This category-to-label mapping exists twice in this file — here, and in the Phase 3 table at lines 220-223 under a "Resulting action" heading. Same rule, two renderings, nothing forcing them to agree.

And it's prose-only: the schema doesn't encode it. I validated a few instances against the committed schema and they all pass —

- `fix_category: environment` with `labels: ["ready-to-code"]` (this table says environment gets no `ready-to-code`)
- `fix_category: infra_flake` with `action: skip` but `ready-to-code` still in `labels`
- the same workspace emitted twice in one run, which produces two issues and two coder runs
- `cycle_ready_to_code: true` on `action: create`, where the post-script never reads it

Phase 3 is about classification; the label mapping belongs only here. Worth deleting the "Resulting action" column up there and cross-referencing this table instead, then encoding the coupling in the schema so a drifting prompt can't silently dispatch the coder on an `environment` failure.

.fullsend/rhdh/schemas/e2e-triage-result.schema.json
```diff
@@ -0,0 +1,135 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "e2e-triage-result.schema.json",
+  "title": "E2E Triage Agent Result",
+  "description": "Structured output from the e2e-triage agent. Classifies nightly E2E failures per workspace and emits issue directives.",
+  "type": "object",
```
gustavolira, line 6: This schema is the contract between a non-deterministic LLM and a script that files public issues, and nothing pins it. `validate-output-schema.sh:57-76` calls `Draft202012Validator(schema).validate(instance)` but never `check_schema()` — I confirmed that renaming `maxLength` to `maxLenght` in here still passes validation and silently drops the constraint, because unknown keywords are legal annotations. So a typo in this file is invisible in CI *and* at runtime.

The repo already has the framework for this: `scripts/tests/` is a pytest suite with committed fixtures, run by `validate-catalog-scripts.yaml`. A ~60-line `test_e2e_triage_schema.py` plus 6-8 fixture files would `check_schema()` this, assert the valid shapes (`create` / `comment` / `skip` / umbrella) pass, and assert the bad ones are rejected. Two of those negatives are findings humans had to catch by eye on this PR — `{"action":"comment","number":null}` and an unconstrained `labels` array — and they'd become regressions CI catches instead.

Cost is adding `jsonschema` to `scripts/requirements-dev.txt` and `.fullsend/rhdh/schemas/**` to the path filter at `validate-catalog-scripts.yaml:9-12`. Not asking for a bash test framework — the repo has never tested a shell script and that's a separate conversation.

.fullsend/rhdh/scripts/post-e2e-triage.sh
```diff
@@ -0,0 +1,290 @@
...
+
+echo "Target branch: $(jq -r '.target_branch // "main"' "${RESULT_FILE}")"
+echo "Workspaces to process: ${WORKSPACE_COUNT}"
+
+# ---------------------------------------------------------------------------
+# 2. Scan agent-result.json for secrets
+# ---------------------------------------------------------------------------
+if ! install_gitleaks; then
+  echo "::error::Failed to install gitleaks — refusing to post without secret scan"
+  exit 1
```
gustavolira, line 163: All six Test plan boxes in the PR description are still unchecked. The end-to-end evidence posted below (issue #2984 → PR #2990 on `release-1.10`) is real and does prove the `create` path plus release-branch handoff — but that run was on 2026-07-28T19:33Z, against commit `d9a742fb`. Five commits have landed since, rewriting +157/-63 across this script and the agent prompt: this very `exit 1` didn't exist during that run, nor did the timeline-API dedup rewrite or the umbrella emit rules.

So what's actually verified is the single-workspace `test_fix` → `create` shape on a 5-revision-old pipeline. Still unproven: dedup emitting `comment` (the search logic was replaced afterwards), `cycle_ready_to_code` (the remove → sleep → add sequence has never executed), the umbrella path, and this fail-closed gitleaks branch. Worth one re-run at the current tip to check boxes 1-4 honestly; 5 and 6 seem reasonable to defer with a note.

.fullsend/rhdh/schemas/e2e-triage-result.schema.json
```diff
@@ -0,0 +1,135 @@
...
+      "required": ["workspace", "fix_category", "tests", "root_cause", "root_cause_slug"],
+      "properties": {
+        "workspace": {
+          "type": "string",
+          "minLength": 1,
+          "maxLength": 128,
+          "description": "Workspace name (e.g., argocd, orchestrator, backstage)."
+        },
+        "fix_category": {
+          "type": "string",
```
gustavolira, line 41: The whole point of the PR is that this agent no longer fixes anything — the harness says so, the prompt says so, the script header says so — but the field is still `fix_category`, and its own description contradicts the name ("Dominant **classification** for failures"). One of the four values, `infra_flake`, explicitly means "no fix, no issue": a fix category that categorically isn't one.

This is the cheapest it will ever be to rename — the field is brand new, and the issue bodies that embed it (`e2e-triage.md:339` writes `fix_category: <CATEGORY>` into the issue) aren't in the wild yet. `failure_category` or just `category`. If a full rename is too much churn, at least expand `FIX_CAT` in the script so the shell variable matches the JSON key it holds.

.fullsend/rhdh/scripts/post-e2e-triage.sh
```diff
@@ -0,0 +1,290 @@
...
+        --title "${ISSUE_TITLE}" \
+        --body "${ISSUE_BODY}" 2>&1)"; then
+        ISSUE_NUMBER="$(echo "${ISSUE_URL}" | grep -o '[0-9]*$')"
+        echo "  Created issue #${ISSUE_NUMBER}: ${ISSUE_URL}"
+        ISSUE_REF="#${ISSUE_NUMBER}"
+
+        # Add labels via labels API. Batch non-deferred labels into one
+        # call, then add ready-to-code LAST so its labeled event is the
+        # final webhook and triggers the coder.
+        NON_DEFERRED="$(jq -c "[.workspaces[$i].issue.labels // [] | .[] | select(. != \"ready-to-code\")]" "${RESULT_FILE}")"
```
gustavolira: `ready-to-code` is the single most load-bearing string in the pipeline — it's what hands work to the coder — and it's embedded five times (209, 210, 217, 242, 244), twice inside escaped jq program text where a typo silently yields "no deferred label" rather than an error. There's already a Configuration block at lines 30-36 for exactly this.

```bash
READY_TO_CODE_LABEL="ready-to-code"
…
OTHER_LABELS="$(jq -c --arg rtc "${READY_TO_CODE_LABEL}" \
  "[.workspaces[$i].issue.labels // [] | .[] | select(. != \$rtc)]" "${RESULT_FILE}")"
```

`--arg` also kills the escaped-quotes-inside-a-double-quoted-jq-program pattern, which is the least reviewable construct in the file.

While here: "deferred" appears nowhere else in the script, schema, or prompt, so `NON_DEFERRED` / `HAS_DEFERRED` only make sense after reading the comment above them — and they're asymmetric anyway (one holds a JSON array, the other the string `"true"`/`"false"`). `OTHER_LABELS` and `WANTS_READY_TO_CODE` say what they hold. Same for `CYCLE` at 227 → `CYCLE_READY_TO_CODE`, and `WS_NAME` at 177, whose `WS_` prefix marked the parallel arrays in the old script that no longer exist here.

.fullsend/rhdh/scripts/post-e2e-triage.sh
```diff
@@ -0,0 +1,290 @@
...
+  echo ""
+  echo "--- Workspace: ${WS_NAME} (${FIX_CAT}) ---"
+
+  ISSUE_REF=""
+
+  case "${ISSUE_ACTION}" in
+    create)
+      echo "  Creating GitHub issue..."
+      ISSUE_TITLE="$(jq -r ".workspaces[$i].issue.title" "${RESULT_FILE}")"
+      ISSUE_BODY="$(jq -r ".workspaces[$i].issue.body" "${RESULT_FILE}")"
```
gustavolira: Line 178 already establishes the right pattern — one `@tsv` pass pulling four fields — and then this arm immediately abandons it and spawns four more `jq` calls (189, 190, 209, 210), with three more in the `comment` arm. The cost isn't CPU, it's that `.workspaces[$i]` gets retyped seven times per iteration, each one a chance to index the wrong record, and the reader has to re-derive "which record am I in" at every line.

Hoisting once per iteration reads much better:

```bash
WS_JSON="$(jq -c ".workspaces[$i]" "${RESULT_FILE}")"
ISSUE_TITLE="$(jq -r '.issue.title' <<<"${WS_JSON}")"
```

Since there are no parallel `WS_*` arrays to correlate by position any more, you could drop the index entirely and iterate `jq -c '.workspaces[]'`.

Related: this `create` arm does six things in 36 lines, and the label orchestration at 206-218 — with its own two-phase ordering constraint — sits 100 lines away from the `add_label`/`remove_label` helpers it belongs next to. Extracting `apply_labels()` and `cycle_ready_to_code()` up beside those helpers would put the ordering rationale next to the code enforcing it, and shrink both case arms to a screenful.

.fullsend/rhdh/agents/e2e-triage.md
```diff
@@ -0,0 +1,570 @@
...
+  Bash(git commit *), Bash(git commit),
+  Bash(gh pr create *), Bash(gh pr edit *), Bash(gh pr merge *),
+  Bash(gh issue create *), Bash(gh issue edit *), Bash(gh issue comment *)
+---
+
+# E2E Nightly Triage Agent
+
+You analyze E2E test failures from the rhdh-plugin-export-overlays nightly CI
+pipeline. You classify failures per workspace and emit issue directives for
+the post-script. You do NOT fix code, create branches, or push — the scaffold
```
gustavolira: "scaffold coder" appears 11 times across five files in this PR, and nothing is called that. The agent it means is `agents/code.md` (`name: code`), wired as `harness/code.yaml`, registered in `config.yaml` under `roles: [triage, coder]`. The only occurrence of "scaffold" anywhere is an unrelated provenance comment in `policies/code.yaml:6`, which is probably where the term leaked in from.

Someone trying to find "the scaffold coder" to understand what happens after triage has nothing to grep for. Calling it "the code agent" (or just "the coder") everywhere makes the handoff traceable.

.fullsend/config.yaml
```diff
@@ -14,8 +14,8 @@ agents:
       source: rhdh/harness/fix.yaml
     - name: review
       source: rhdh/harness/review.yaml
-    - name: e2e-fix
-      source: rhdh/harness/e2e-fix.yaml
+    - name: e2e-triage
+      source: rhdh/harness/e2e-triage.yaml
```
gustavolira, line 18: nit: `docs/fullsend.md` is the repo's written fullsend doc — it enumerates the enabled agents at :9-16 and the configuration files at :182-189 — and this PR adds a fourth repo-owned agent, a new label trigger, a new workflow, and a new post-script without touching it. It's already stale (it never listed `e2e-fix` either, and :71 still says "one custom script"), so this doesn't make things worse — but rewriting the whole system is a cheap moment to add the row.

.fullsend/rhdh/agents/e2e-triage.md
subhashkhileri: Fixed in 3410a1bc — consolidated all three templates (per-workspace, umbrella, product_bug) into a single unified template. `**Target branch:**` now appears once in the template and applies to all variants including product_bug and umbrella per-workspace remediation sections. Also removed redundant category→action mapping from Phase 3, duplicate field rules from Phase 6, and duplicate prescriptive-writing example from Constraints. Net: -72 lines, zero information lost.

.fullsend/rhdh/agents/e2e-triage.md
subhashkhileri: Addressed in the same commit (3410a1bc) — see reply above. All three templates are now one unified template, so `Target branch` can't be dropped by any variant.

.fullsend/rhdh/schemas/e2e-triage-result.schema.json
subhashkhileri: Fixed in 35f162f5 — `labels.items` is now `{"enum": ["e2e-failure", "ready-to-code"]}`. Agent can no longer emit `e2e-triage` or any other label that could trigger a dispatch loop.

.fullsend/rhdh/schemas/e2e-triage-result.schema.json
subhashkhileri: Fixed in 35f162f5 — changed `number` type from `["integer", "null"]` to `"integer"`. The field is optional (only required when `action: "comment"` via the if/then conditional), so `create` and `skip` can omit it entirely. Also updated the agent's output template to not show `"number": null`.

.fullsend/rhdh/scripts/post-e2e-triage.sh
subhashkhileri: Fixed in 35f162f5 — pinned to `linux_x64` explicitly and dropped the arch detection switch. CI runners are always `ubuntu-latest` (x64), so the arm64 branch was dead code. The checksum now matches the only URL we'll ever hit.

.fullsend/rhdh/scripts/post-e2e-triage.sh
subhashkhileri: Fixed in 35f162f5 — stderr now goes to a temp file (like `add_label` does), issue number extracted via `${ISSUE_URL##*/}` (same pattern as line 52), and validated with `^[0-9]+$` before use in API paths.

.fullsend/rhdh/scripts/post-e2e-triage.sh
subhashkhileri: Fixed in 35f162f5 — added `IFS=$'\t'` before `read -r`. The `@tsv` output is tab-delimited, so splitting only on tabs prevents field shift if workspace names or slugs contain hyphens or other characters.

.fullsend/rhdh/agents/e2e-triage.md
subhashkhileri: Fixed in 35f162f5 — split tracking into separate lines per key:
```
`fullsend-tracking: workspace=<name>`
`fullsend-tracking: root-cause=<slug>`
`fullsend-tracking: branch=<branch>`
```
Each key is now independently searchable via GitHub's `in:body` exact-phrase matching. The per-workspace search (`workspace=X`) and umbrella search (`root-cause=Y`) both match as prefix phrases on their own line.

.fullsend/rhdh/agents/e2e-triage.md
subhashkhileri: Fixed in 35f162f5 — two issues found and fixed:

1. **Escaped `\%`** — `${text//%0A/}` never actually matched because bash interprets the leading `%` after `//` as an end-anchor. All `%0A`/`%0D` stripping was a no-op. Fixed with `${text//\%0A/}`.
2. **Loop until stable** — wraps all replacements in a `while [[ "${text}" != "${prev}" ]]` loop to catch reconstitution from crafted strings like `%%0A0A`.

Tested all bypass vectors from the review — all pass now.

.fullsend/rhdh/policies/e2e-triage.yaml
subhashkhileri: Fixed in 35f162f5 — added `Edit, Write, MultiEdit` to `disallowedTools` in the agent frontmatter. The read-only constraint claimed in the Constraints section is now enforced.

subhashkhileri: Re: curl in policy — intentional. The `github_api` network policy is `access: read-only`, so curl with GH_TOKEN can only read, not write. No security risk. Keeping it for flexibility.

.fullsend/rhdh/scripts/post-e2e-triage.sh
subhashkhileri: Fixed in e53cfc83 — `remove_label` now returns meaningful exit status: 0 on success or 404 (label wasn't there), non-zero on real API errors. The cycle block only proceeds with `add_label` if removal succeeded, warns otherwise.

.fullsend/rhdh/scripts/post-e2e-triage.sh
subhashkhileri: Skipped — agent Phase 4 already runs deterministic `gh api search/issues` with structured tracking lines. The #16 fix (tracking lines split onto separate lines) made those searches reliable. Low risk, not worth duplicating logic in post-script.

.fullsend/rhdh/scripts/post-e2e-triage.sh
subhashkhileri: Fixed in e53cfc83 — summary now built with real newlines (\$'\\n') and `printf '%s'` instead of escaped \\n + `printf '%b'`. Agent text containing \\d, \\n, \\t is no longer mangled.

.fullsend/rhdh/harness/e2e-triage.yaml
subhashkhileri: Fixed in e53cfc83 — added `GITHUB_ISSUE_URL` to `forge.github.env.runner`. Forge semantics are merge (not replace) per ADR-0045/0055, so the top-level entry would survive — but this matches the fullsend scaffold's own `triage.yaml` convention of putting forge-specific webhook vars in the forge block.

.fullsend/rhdh/policies/e2e-triage.yaml
subhashkhileri: Confirmed by operational evidence — all four harness files (`code.yaml`, `fix.yaml`, `review.yaml`, `e2e-triage.yaml`) use the same image (`ghcr.io/redhat-developer/rhdh-fullsend-code:latest`) and all include the `playwright-trace` skill. The code and fix agents run successfully in production. No code change needed.

.fullsend/rhdh/agents/e2e-triage.md
subhashkhileri: Intentional — JIRA filing was deliberately dropped from fullsend scope. Not creating JIRA issues through fullsend is a design decision.

.fullsend/rhdh/scripts/post-e2e-triage.sh
subhashkhileri: Fixed in e53cfc83 — added `Root Cause` column to the summary table, populated from `root_cause_slug`. Rows now read `| orchestrator | infra_flake | ocp-timeout | 3 | — |` instead of having no context for flake entries.

.fullsend/rhdh/harness/e2e-triage.yaml
subhashkhileri: Fixed in e53cfc83 — re-added `rhdh-toolchain.env` mount. Sets `COREPACK_HOME=/tmp/corepack` (yarn needs a writable dir) and `OPENSPEC_TELEMETRY=0` (suppresses DENIED requests to `edge.openspec.dev` which isn't in the triage policy).

.fullsend/rhdh/policies/e2e-triage.yaml
subhashkhileri: Fixed in e53cfc83 — updated provenance from `Based on rhdh/policies/e2e-fix.yaml` (deleted by this PR) to `forked-from: rhdh/policies/code.yaml`.

.github/workflows/e2e-triage-agent.yaml
subhashkhileri: Fixed in e53cfc83 — added `/.github/workflows/e2e-triage-agent.yaml` to CODEOWNERS with the same reviewers as `.fullsend/` and `fullsend.yaml`.

.fullsend/rhdh/agents/e2e-triage.md
subhashkhileri: Skipped — the overloading works in practice (coder reads the issue body which lists all workspaces, post-script handles it correctly). Adding `umbrella`/`covers` schema fields would be ~30 lines of churn across 3 files for minimal practical benefit.

.fullsend/rhdh/schemas/e2e-triage-result.schema.json
subhashkhileri: Fixed in 84f79e87 — raised `maxItems` for workspaces from 30 to 50 and for tests from 50 to 200. Prevents schema rejection on mass failures.

AGENTS.md
subhashkhileri: Skipped — triage agent is read-only (`Edit`, `Write`, `MultiEdit` already in `disallowedTools`). The prohibited modifications list is for the code agent, not relevant here.

.fullsend/rhdh/agents/e2e-triage.md
subhashkhileri: Already fixed in 3410a1bc — the "Resulting action" column was removed from Phase 3 during the template consolidation. Single source of truth for category→label mapping is now Phase 5.

.fullsend/rhdh/schemas/e2e-triage-result.schema.json
subhashkhileri: Acknowledged — valid point about `check_schema()` and negative test cases. Adding a pytest suite is separate PR scope. Not blocking.

.fullsend/rhdh/scripts/post-e2e-triage.sh
subhashkhileri: Acknowledged — will do a re-run at current tip before merge to verify the changes since `d9a742fb` (gitleaks exit, sanitize_for_gha, tracking lines, remove_label, summary table).

.fullsend/rhdh/schemas/e2e-triage-result.schema.json
subhashkhileri: Skipped — valid naming point but touches schema + agent + post-script. Moderate churn for a cosmetic rename. Can revisit in a follow-up if needed.

.fullsend/rhdh/scripts/post-e2e-triage.sh
subhashkhileri: Skipped — valid but moderate churn (constant extraction + `jq --arg` + variable renames). Can revisit in a follow-up cleanup.

.fullsend/rhdh/scripts/post-e2e-triage.sh
subhashkhileri: Fixed in 84f79e87 — hoisted `WS_JSON=$(jq -c ".workspaces[$i]" ...)` once per iteration. All subsequent jq calls read from the variable instead of re-indexing the file. Single index, single source of truth.

.fullsend/rhdh/agents/e2e-triage.md
subhashkhileri: Fixed in 84f79e87 — renamed "scaffold coder" → "code agent" across all 4 files (8 occurrences). Maps to `code.md`/`code.yaml` and is now traceable via grep.

.fullsend/config.yaml
subhashkhileri: Fixed in 84f79e87 — updated `docs/fullsend.md` with e2e-triage agent rows in both tables, `e2e-triage-agent.yaml` in the config files table, and CODEOWNERS section.

PR #2991, opened 2026-07-28
feat(metadata): validate appConfigExamples against plugin config schemas (RHIDP-13509)

gustavolira:
## What

Extends the `appConfigExamples` check so it validates each example against the
plugin's own config schema, not just that an example exists ([RHIDP-13509](https://redhat.atlassian.net/browse/RHIDP-13509)).

Doing that means compiling `config.d.ts` declarations, so the validator moves
from Python to TypeScript and reuses `@backstage/config-loader` — the same
machinery Backstage uses at runtime.

## Commits

1. **Port to TypeScript**, no behaviour change. Verified by a single property:
   output is byte-identical to the Python script across all 178 metadata files,
   with the same exit code. Re-checked after every subsequent commit.
2. **Semantic validation** against the published package's schema.
3. **Type guards** replacing three assertions.
4. **Review fixes** — see below.

## What it catches, and what it doesn't

Verified against the real compiler rather than assumed.

**Caught:** non-coercible scalars, wrong nesting, bad enum values, missing
required properties.

**Not caught, and documented as such:**
- Coercible scalars. Ajv runs with `coerceTypes: true`, so `port: "8080"`
  against a declared number passes.
- Undeclared keys. 72 of 178 files carry a `dynamicPlugins` block that belongs
  to no plugin schema, and 65 contain nothing else, so rejecting undeclared keys
  would fail them all.
- Packages whose `config.d.ts` imports from their dependencies — `npm pack`
  brings no `node_modules`. 6 of a 29-package sample.

Every schema run prints a validated/mismatched/no-schema/unavailable tally and
says so outright when nothing was validated, so the gate cannot be green merely
because it is inert.

## Findings on the current tree

A full sweep surfaced two genuine mismatches:

- `analytics-module-segment` — `/app/analytics/segment/testMode` fails the
  schema's `anyOf`
- `dynatrace-dql` — `/dynatrace/environments` missing the required
  `platformToken`

## Review

Three independent reviews ran against this branch before it was opened; the
fourth commit addresses what they found. Highlights worth a reviewer's
attention:

- The PR step runs `--since`, which by construction sees no metadata when a PR
  only touches this directory — so the semantic path would have shipped without
  executing once. A smoke step now exercises it against known metadata.
- `packageName`/`version` reach `npm pack` as argv and come from metadata a fork
  PR controls. Both are validated now. The test for it caught that the first
  regex used `[a-z0-9-~]`, where `9-~` is a range covering most of printable
  ASCII — it let through exactly the input it was meant to block.
- `splitSchemaErrors` never split anything: config-loader joins violations with
  `; ` on one line, so the wall of text it claimed to prevent was what shipped.

## Testing

62 tests, none touching the network. The semantic layer is exercised through
`loadConfigSchema({ serialized })`, which builds a real Backstage schema in
memory, so the suite tests the actual validator rather than a stand-in.


scripts/validate-app-config-examples/README.md
```diff
@@ -0,0 +1,110 @@
...
+**Not caught:**
+
+- **coercible scalars.** `@backstage/config-loader` builds Ajv with
+  `coerceTypes: true`, so `port: "8080"` against a declared number passes. This
+  is one of the more common real app-config mistakes, and this check does not
+  see it.
+- **undeclared keys.** Examples legitimately carry RHDH wiring that belongs to
+  no plugin schema — 72 of 178 metadata files include a `dynamicPlugins` block
+  and 65 contain nothing else. Rejecting undeclared keys would fail all of them,
+  so a typo'd key name passes silently.
```
gashcrumb, line 81: Worth calling out so we don’t over-read what this gate promises.

The check still allows coercible scalars (e.g. `port: "8080"`), undeclared keys (including `dynamicPlugins`-only examples), and typo’d key names. That’s documented and reasonable given the catalogue shape — but it means this won’t catch a fair share of “paste into `dynamic-plugins.yaml` and get a runtime error” cases.

Two questions:
1. Is this README the intended long-term contract for that boundary, or do we want a follow-up (e.g. optional `noUndeclaredProperties` for packages that don’t rely on the `dynamicPlugins` wrapper pattern)?
2. For the two known mismatches (`analytics-module-segment`, `dynatrace-dql`), should we track a follow-up to fix the metadata so the backlog isn’t warn-only forever?

scripts/validate-app-config-examples/package.json
```diff
@@ -0,0 +1,29 @@
...
+    "node": ">=24",
+    "yarn": ">=4"
+  },
+  "packageManager": "yarn@4.12.0",
+  "scripts": {
+    "build": "esbuild src/validate.ts --bundle --platform=node --format=esm --packages=external --outfile=dist/validate.mjs",
+    "validate": "yarn build && node dist/validate.mjs",
+    "test": "esbuild src/*.test.ts --bundle --platform=node --format=esm --packages=external --outdir=dist-tests --out-extension:.js=.mjs && node --test \"dist-tests/*.test.mjs\"",
+    "tsc:check": "tsc --noEmit",
+    "check": "yarn tsc:check && yarn test"
```
gashcrumb: This package has `tsc` + tests but no eslint/prettier, unlike the other standalone TypeScript projects in the repo. It also uses single quotes vs the double-quote convention elsewhere.

Not blocking if you want to keep the tool lean, but adding the same `lint:check` / `prettier:check` pattern (and wiring them into `yarn check`) would keep style enforcement consistent and avoid drift as more people touch this.

gashcrumb (commented):
A couple of non-blocking questions on the semantic-check contract and tooling consistency — see inline.

scripts/validate-app-config-examples/README.md
gustavolira: Good calls — filed a Jira for each.

Hadn't thought of making `noUndeclaredProperties` conditional; I'd dropped it as all-or-nothing because of the wrapper examples. Measured it: 70 files have example content with no `dynamicPlugins` key, so ~40% of the catalogue could get typo detection. Filed as [RHIDP-15902](https://issues.redhat.com/browse/RHIDP-15902).

On the mismatches, agreed — warn-only was meant to stop a pre-existing backlog wedging the sweep, not to make it permanent. Filed as [RHIDP-15903](https://issues.redhat.com/browse/RHIDP-15903). `dynatrace-dql` is Dynatrace's so that half needs them; `analytics-module-segment` we can just fix.

scripts/validate-app-config-examples/package.json
gustavolira: Done — `lint:check` / `lint:fix` / `prettier:check` / `prettier:fix`, and `yarn check` now runs tsc + eslint + prettier + tests. Prettier's defaults sorted out the quotes too (9 files reformatted).

eslint uses a minimal `typescript-eslint` config rather than the shared `@red-hat-developer-hub/e2e-test-utils/eslint` one — that pulls in `eslint-plugin-playwright` and this is a CLI.

Also opened https://github.com/redhat-developer/rhdh-plugin-export-overlays/pull/3003 to do the same to `smoke-tests-native/`, which had the same gap. Seemed odd to fix only the new one.

gashcrumb (approved):
Thanks for the follow-ups — lint/prettier looks good, and RHIDP-15902 / RHIDP-15903 cover the remaining gaps. LGTM.

PR #2996, opened 2026-07-29
fix(intelligent-assistant-e2e): disable Segment telemetry plugin

HusneShabbir:
## Summary
- Disable `analytics-provider-segment` in Intelligent Assistant e2e dynamic plugins config (PR/local and nightly) to avoid outbound analytics requests that slow RHDH
- Always wire `dynamicPlugins` in the deploy helper so the Segment override is applied outside nightly mode too
- Ports the previous lightspeed e2e telemetry fix to the renamed `intelligent-assistant` workspace

**Resolves:**
https://redhat.atlassian.net/browse/RHIDP-15506

## Test plan
- [ ] Confirm `dynamic-plugins.yaml` lists Segment as `disabled: true` plus both Intelligent Assistant packages
- [ ] Confirm nightly config disables Segment only (no duplicate workspace OCI entries)
- [ ] Run Intelligent Assistant e2e (or UI mode with existing deployment) and verify plugins still load without Segment telemetry traffic
- [ ] Optionally comment `/test` on the PR if CI e2e coverage for this workspace is desired


workspaces/intelligent-assistant/e2e-tests/tests/config/dynamic-plugins.yaml
```diff
@@ -0,0 +1,11 @@
+# Explicit OCI refs (not ./dynamic-plugins/dist wrappers). Listing packages here
+# skips metadata auto-generation, so IA must be included for PR/local injection +
+# OCI resolution, and for nightly until the catalog index ships IA (1.11 still
+# has lightspeed). Segment is disabled to avoid outbound analytics that slow RHDH.
+plugins:
+  - package: oci://ghcr.io/redhat-developer/rhdh-plugin-export-overlays/backstage-community-plugin-analytics-provider-segment:bs_1.52.0__1.30.0
```
gashcrumb, line 6: would it make sense to add a `disablePlugin` function to rhdh-e2e-test-utils instead of this workspace specific fix?  Also I think this might break currently if/when the catalog points to the build on quay.

PR #2999, opened 2026-07-29
test(extensions): update e2e to run for GA plugins only

jrichter1:
adding config that was removed from DPDY, update tests to use the currently available plugins


sanketpathak:
Verified locally 
Screenshot 2026-07-30 at 7 38 04 PM

/lgtm

